PR #1, opened 2015-12-29
Extract apiclient from dmutils

allait:
### Copy apiclient modules and tests from digitalmarketplace-utils
(https://github.com/alphagov/digitalmarketplace-utils/tree/63531bd4d2c1bed8774fef04aaf3ac8f71d341db)

### Add dmutils.audit and audit types tests from digitalmarketplace-utils
Audit event types enum is used only for the API client calls, so it
makes sense to group it with the API client module.

### Rewrite apiclient tests to avoid dmutils.request_id dependency
Tests for request headers were using the `dmutils.request_id` module
to set up the Request class with `.request_id` attribute. Since the
test is supposed to check that API client is setting the header on
the API request, we can configure the Flask request object directly
instead of relying on `request_id.init_app`.

### Use dmapiclient package name instead of dmutils namespacing
Using dmutils namespace with namespace_packages requires removing
all code from `dmutils.__init__`, which is currently acting as an
import proxy for third-party packages. Removing `__init__.py` will
break app imports, which is the main reason for package namespacing.

Instead, we create a new package name `dmapiclient`, which allows us
to simplify the directory structure by removing the nested `apiclient`
package. The existing `dmutils.apiclient` import path can be preserved
by adding `import dmapiclient as apiclient` to `dmutils/__init__.py`.


README.md
```diff
@@ -0,0 +1,9 @@
+Digital Marketplace API client
+=========================
+
+[![Coverage Status](https://coveralls.io/repos/alphagov/digitalmarketplace-apiclient/badge.svg?branch=master&service=github)](https://coveralls.io/github/alphagov/digitalmarketplace-apiclient?branch=master)
+[![Requirements Status](https://requires.io/github/alphagov/digitalmarketplace-apiclient/requirements.svg?branch=master)](https://requires.io/github/alphagov/digitalmarketplace-apiclient/requirements/?branch=master)
+
+## What's in here?
+
+Digital Marketplace Data and Search API clients
```
robyoung: This could do with a bit more info. Maybe just an example and links to the other relevant repos.

scripts/push-tag.sh
```diff
@@ -0,0 +1,55 @@
...
+    git push origin --tags
+}
+
+function main(){
+    get_sha
+    get_version
+    push_tag_or_die
+}
+
+main
```
robyoung: Copying this between the repos makes me twitchy but I suppose it doesn't change often.

robyoung:
Other than the README point this looks ace :+1: :pray:

scripts/push-tag.sh
allait: Yeah, I'm not really sure what to do about it either. We could move it to the Jenkins job builder, since it's the only place that's actually using it now.

allait:
Expanded the README a bit and added `CHANGELOG.md` for 1.0.0

PR #2, opened 2015-12-30
Add 504 errors to be retried together with 503 in iter_ methods

allait:
503 status is returned for all connection or response-related issues,
however it doesn't include the 504 responses returned by AWS ELBs for
occastional request failures. Since 504 requests should be retried
for GET requests the same way connection errors are, 504 is added
to the list of statuses that return the HTTP503Error exception.

Since HTTP503Error is a misleading name now, the exception is renamed
to `HTTPTemporaryError`.

This should fix the issues with the index-services job failing due
to 504 responses when listing the services.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '1.0.0'
+__version__ = '1.0.1'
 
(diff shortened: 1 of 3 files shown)


robyoung:
:+1:

PR #3, opened 2016-01-04
Add parameter to order audit events

quis:
Implements: alphagov/digitalmarketplace-api#315

Bumps version to: **`1.1.0`**

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -1289,7 +1289,7 @@ class TestDataApiClient(object):
 
     def test_find_audit_events_with_all_params(self, data_client, rmock):
         rmock.get(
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -1300,7 +1300,8 @@ def test_find_audit_events_with_all_params(self, data_client, rmock):
             acknowledged='all',
             audit_date='2010-01-01',
             object_type='foo',
-            object_id=123)
+            object_id=123,
+            latest_first='True')
```
allait: This probably should be a boolean `True`, not a string. At least that's how I'd expect the method to be called.

quis: @allait Don’t know how those quotes snuck in there. Rebased out now.

allait:
:+1:

PR #4, opened 2016-01-19
Add audit types for deleting and uploading countersigned agreements

carolinegreen:


TheDoubleK:
This PR is weird - looks like it is already merged but still open?

Anyway, I'm wondering if there is really a requirement to be able to delete agreements? With all other documents updating a new version adds a timestamp but we keep the previous version with an earlier timestamp.

I'm pretty sure if the agreement gets countersigned then CCS deleting it from the server doesn't legally constitute them withdrawing the agreement. So I'm not sure we should be deleting anything at all.

carolinegreen:
I think part of it has been merged but I'm not really sure what has happened there.

It is a requirement for the ccs sourcing admin to be able to remove agreements, but since a copy is always taken at the time the agreement is uploaded the actual document is never completely deleted. 

If they shouldn't be able to delete agreements then surely they shouldn't be able to replace agreements either, and the whole story would be rendered pointless.

TheDoubleK:
You could still have upload and replace files as a story without any way to delete them altogether. I've seen it's a requirement in the story, but it still doesn't make sense to me why CCS would ever need to delete an agreement  - could be in case they upload the wrong one I guess?

Anyway, :+1: for the version bump, which seems to be the only change left to go in here now.

carolinegreen:
Deleting is doing almost the same thing as a replacement, as there is still a copy of the original file in both cases, there simply wouldn't be an updated file shown and it would appear on the admin app as if there were no files.  I think that's fine.

PR #5, opened 2016-01-25
Add method to update status of a draft

TheDoubleK:
This adds a method to access the new API endpoint defined here:  https://github.com/alphagov/digitalmarketplace-api/pull/333

This PR need to be merged before the API PR, because it needs to use the new audit type defined here.


robyoung:
:+1:

PR #7, opened 2016-02-02
Role-based authentication

TheDoubleK:
As per @robyoung's comment here:
https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/210#discussion-diff-51412643R264

Rather than having a flag for `supplier` we can pass the desired user role in for authentication.


TheDoubleK:
This isn't quite what we need after all...

PR #8, opened 2016-02-02
Role based authentication

TheDoubleK:
As per @robyoung's comment here:
alphagov/digitalmarketplace-buyer-frontend#210 (diff)

Rather than having a flag for supplier we can pass the desired user role in for authentication.

**QUESTION:** Do we actually need this `role` parameter at all now?  The login route in the buyer app is for logging in buyers or suppliers, so when do we actually need to authenticate against a specific role?  It doesn't hurt to have it here, but I'm not sure we need it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@ PR: [#8](https://github.com/alphagov/digitalmarketplace-apiclient/pull/8)
 ### What changed
 
 The `authenticate_user` method used to take a boolean flag to indicate if the user is a supplier user.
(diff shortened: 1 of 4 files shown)


robyoung:
Yea, I kinda agree with the premise of your question. I don't think we need to be checking the role in this function. If we keep this about just authenticating we can do any role checking, if needed, afterwards.

TheDoubleK:
OK great - I've removed the role checking from this method.

robyoung:
:+1:

PR #10, opened 2016-02-10
Add stubs for use in tests

robyoung:
Now that we have complex API interactions in the buyers app it would be useful to have common API response factories somewhere they can be shared between the frontend apps.

We have decided not to look at Factory Boy right now as our use case is pretty simple.

## Example use
```python
api_stubs.framework(
  slug="my-framework",
  lots=[
    api_stubs.lot(slug="my-lot", allows_brief=False)
  ]
)
```


TheDoubleK:
:+1:

PR #11, opened 2016-02-11
Add create_brief_response audit event type

allait:
Used when supplier submits a brief response. Since responses are
not editable at the moment we don't need any other event types.


TheDoubleK:
:+1:

PR #13, opened 2016-02-19
Add delete brief method and add timestamps to brief stubs

TheDoubleK:
The brief summary page looks at the last updated time, and will also show the published time for live briefs, so it's useful to have the timestamps in the `brief()` api stub helper.

Also adds a title for the brief, also for testing the brief summary page.

This also now adds an apiclient method for the delete_brief API endpoint that now exists here: https://github.com/alphagov/digitalmarketplace-api/pull/351
(But this will need to be merged first, because the API needs the new audit type introduced here.)

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '2.3.1'
+__version__ = '2.3.2'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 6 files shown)


pcraig3:
Made a comment about a comment.

But looks good :+1:

TheDoubleK:
I'm going to merge it and fix the `s` some time soon, because I can't easily do it on my phone right now.

PR #14, opened 2016-02-22
Brief responses apiclient methods

allait:
### Always send "updated_by" field instead of "update_details"
Now that API supports both "updated_by" and "update_details" keys
we can switch the apiclient to always send the same payload.

Adds helper methods to add user info to the request data for post,
put and delete requests.

Changes data API client methods to use the new helpers so that update
data is always sent in the same format.

### Add brief-responses data API client methods
Adds methods for creating, reading and listing brief responses.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 Records breaking changes from major version bumps
 
+## 3.0.0
+
+PR: [#14](https://github.com/alphagov/digitalmarketplace-apiclient/pull/14)
+
+### What changed
+
+All POST, PUT and DELETE methods are now sending `"updated_by"` payload instead of `"update_details"`.
+The new payload should be supported by the data API for all methods.
+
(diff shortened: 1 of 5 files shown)


pcraig3:
Is the idea to do away with `update_details` entirely, or is there still some reason for them?

allait:
@pcraig3 we don't really need update details since we're only sending `"updated_by"` ever since we've dropped `"update_reason"`, so the idea is to drop the nested dictionary in favour of just `"updated_by"`.

That said, similar to the change in the API this is mostly about making sure that all methods are sending/accepting the same consistent format, I'm less attached to what that format is, so if everyone thinks `"update_details"` is better we can keep that.

pcraig3:
Right, thought so.  

Seems like the ambiguity of `update_details` didn't give us much that was useful.

`updated_reason` is straightforward, so let's keep that.

:+1:

PR #17, opened 2016-03-02
Unset framework agreement returned

allait:
### Add unset_framework_agreement_returned method
Sets supplier framework `agreementReturned` flag to `False`.
Required to remove outdated signed framework agreements after the
original agreement document has been changed and requires a new
signature.


TheDoubleK:
Well, that's pretty horrible.
:+1:

PR #18, opened 2016-03-11
[#114474769] Add client method for new check-buyer-email API route

TheDoubleK:
Client method for the new API endpoint here: https://github.com/alphagov/digitalmarketplace-api/pull/363

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -330,7 +330,7 @@ class DataAPIClient(BaseAPIClient):
             "/users/export/{}".format(framework_slug)
         )
 
-    def email_address_has_valid_buyer_domain(self, email_address):
+    def is_email_address_with_valid_buyer_domain(self, email_address):
         return self._get(
             "/users/check-buyer-email", params={'email_address': email_address}
         )['valid']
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -330,6 +330,11 @@ def export_users(self, framework_slug):
             "/users/export/{}".format(framework_slug)
         )
 
+    def email_address_has_valid_buyer_domain(self, email_address):
```
robyoung: We talked about naming these `is_something_something` so that it's immediately obvious which methods should return booleans.

TheDoubleK: We did, didn't we, and then I got sidetracked by a long discussion about the spacing of bullet points on a page that no-one should ever really see :/

allait: `is_/has_` prefix convention works really well on property methods, eg for things like `email_address.has_valid_buyer_domain`. I think it's less useful when the owner of the method is not the subject of the "question" in the method name.

But it's just a thought, I don't really mind either way.

TheDoubleK: I know what you mean, but I think there is some use in doing something to distinguish API client methods that are returning a boolean rather than a JSON response (there is another one coming soon to check if a supplier is eligible to apply to a brief once this gets merged: https://github.com/alphagov/digitalmarketplace-api/pull/359).  Whether this is the best naming convention I'm not sure, but it's something at least...

allait:
:+1:

PR #19, opened 2016-03-15
Add is_supplier_eligible_for_brief DataAPIClient method

allait:
### Add is_supplier_eligible_for_brief DataAPIClient method
Calls the related API endpoint and returns True if supplier is
eligible to respond to a brief or False otherwise.

Can be used to check whether "Apply" buttons should be displayed
or show an error page to the user, doesn't store anything in the
API.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.4.1'
+__version__ = '3.5.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


TheDoubleK:
:+1:

robyoung:
This needs updating after https://github.com/alphagov/digitalmarketplace-api/pull/359 please don't merge. I'm working on it.

allait:
Was about to update this with new URLs, but turns out Rob already did. Nice! :smile::100:

PR #20, opened 2016-03-16
Additional parameters for find_briefs method (status, lot and framework)

TheDoubleK:
Update to API client method now that the `find_briefs` endpoint can filter on status, lot and framework.

Also adds an iterator method for getting briefs, as brief fetches are paginated.


robyoung:
:+1:

PR #22, opened 2016-03-23
Update brief stub

robyoung:
Includes fields that are autogenerated when a brief is published.


allait:
With the fake date set to the near future, it's easy to accidentally start relying on it in the application code/tests for the next week or so. I'd rather have a date in the past, but since `createdAt` is similar, this might be fine too :+1:

PR #23, opened 2016-03-30
Add audit event for create_framework

robyoung:
I have chosen to go with `create_framework` rather than `framework_create` because that is how most of the other audit events are named.


pcraig3:
:+1:

PR #25, opened 2016-06-09
find_services: add lot parameter

risicle:
which was already supported by API but not exposed here. Bump to 3.8.0

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '3.7.0'
+__version__ = '3.8.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


allait:
Did we check that none of the callers pass positional arguments after `frameworks`? Otherwise this is a breaking change.

risicle:
Hmm good point.

risicle:
/me hates the positional/keyword arg clash situation. wish they were mutually exclusive or something

risicle:
Hold on, I can just move the kwarg to the end of the call signature. And here was me about to do an audit of all call sites. Will rehash.

TheDoubleK:
:+1: good to go.

risicle:
wheee!

PR #26, opened 2016-06-20
Added and updated methods for signerDetails and countersigned field for framework interest

idavidmcdonald:
Updated register_framework_agreement_returned method to take an optional signer_details parameter. If parameter is provided then signer_details will be set/updated. If parameter is not provided then signer_details is left unchanged. Tests added for both of these cases.

Added register_framework_agreement_countersigned method with corresponding test.

We decided not to alter the signer_details field by default when calling `register_framework_agreement_returned` method because silently overwriting it by default would be surprising behaviour.

Removed previous unset_framework_agreement_returned method which is no longer being used and therefore not needed. Therefore bumped version to 4.0.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 Records breaking changes from major version bumps
 
+## 4.0.0
+
+PR: [#26](https://github.com/alphagov/digitalmarketplace-apiclient/pull/26)
+
+### What changed
+
+Removed the `unset_framework_agreement_returned` method.
+
+This shouldn't require any changes as this method is only used in at most one old script and should 
+never need to be used again.
+
 ## 3.0.0
 
 PR: [#14](https://github.com/alphagov/digitalmarketplace-apiclient/pull/14)
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -195,34 +195,12 @@ class DataAPIClient(BaseAPIClient):
             user=user,
         )
 
-    def unset_framework_agreement_returned(self, supplier_id, framework_slug, user):
-        return self._post_with_updated_by(
-            "/suppliers/{}/frameworks/{}".format(
-                supplier_id, framework_slug),
-            data={
-                "frameworkInterest": {
-                    "agreementReturned": False,
-                },
-            },
-            user=user,
-        )
-
     def register_framework_agreement_countersigned(self, supplier_id, framework_slug, user):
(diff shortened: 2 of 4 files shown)


idavidmcdonald:
Related to the [API pull request](https://github.com/alphagov/digitalmarketplace-api/pull/408)

dmapiclient/data.py
```diff
@@ -181,22 +181,48 @@ def set_framework_result(self, supplier_id, framework_slug, is_on_framework, use
...
+            framework_interest_dict['signerDetails'] = signer_details
+
+        return self._post_with_updated_by(
+            "/suppliers/{}/frameworks/{}".format(
+                supplier_id, framework_slug),
+            data={"frameworkInterest": framework_interest_dict},
+            user=user,
+        )
+
+    def unset_framework_agreement_returned(self, supplier_id, framework_slug, user):
```
TheDoubleK: We should lose the unset methods for both dates. This will be a breaking change, requiring a major version number bump and an entry in the CHANGELOG detailing the breaking change. (Although the unset agreement returned at is only used in one script, and should never need to be used again, so this won't have any knock-on effect for our apps.)

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
...
 
+## 4.0.0
+
+PR: [#26](https://github.com/alphagov/digitalmarketplace-apiclient/pull/26)
+
+### What changed
+
+Removed the `unset_framework_agreement_returned` method.
+
+This shouldn't require any changes as this method is only used in at most one old script and should
```
TheDoubleK: In fact, only here: https://github.com/alphagov/digitalmarketplace-scripts/blob/master/scripts/clear-signed-agreements.py#L64

tests/test_apiclient.py
```diff
@@ -1026,31 +1026,70 @@ def test_set_framework_result(self, data_client, rmock):
             'updated_by': 'user'
         }
 
-    def test_register_framework_agreement_returned(self, data_client, rmock):
+    def test_register_framework_agreement_returned_with_signer_details(self, data_client, rmock):
         rmock.post(
             "http://baseurl/suppliers/123/frameworks/g-cloud-7",
-            json={"frameworkInterest": {"agreementReturned": True}},
+            json={
```
TheDoubleK: It doesn't look like this `json` variable is used at all?

TheDoubleK:
Looks good :+1:

PR #27, opened 2016-07-01
`signer_details` -> `agreement_details` + un-remove unsetting framework agreements

pcraig3:
####  Rename `signer_details` to `agreement_details`

We are expanding the scope of what the `*_details` JSON field will store to do with signing framework agreements.  (For example, not all stored values will be about the supplier who signs it -- we are also planning to save the name/role of the countersigner.)

With this in mind, we are generalising the name of the column in the database.  This is easy to do now because no one is using it yet.

#### Add back a way to unset framework agreements
We used them for DOS, so @allait says keep them in.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '4.0.0'
+__version__ = '4.1.0'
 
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
👍

PR #28, opened 2016-07-11
Return framework agreement and update `agreementDetails`

idavidmcdonald:
Added `update_supplier_framework_agreement_details` function to set and update keys in `agreementDetails`

Changed `register_framework_agreement_returned` from taking an optional `agreement_details` parameter to taking an optional `uploader_user_id` paramater. This is because when returning an agreement (that has a framework agreement version), we expect only an `uploader_user_id` instead 
of a flexible dictionary and have made this method stricter to enforce/strongly encourage this. There should currently be no usage of this in our applications as this feature is not yet released and is being tweaked since the previous related [pull request](https://github.com/alphagov/digitalmarketplace-apiclient/pull/27).

Due to the above breaking change, we bump to 5.0.0.

This PR relies on the following API [pull request](https://github.com/alphagov/digitalmarketplace-api/pull/416).

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -181,12 +181,12 @@ class DataAPIClient(BaseAPIClient):
             user=user,
         )
 
-    def register_framework_agreement_returned(self, supplier_id, framework_slug, user, agreement_details=None):
+    def register_framework_agreement_returned(self, supplier_id, framework_slug, user, uploader_user_id=None):
         framework_interest_dict = {
             "agreementReturned": True,
         }
-        if agreement_details is not None:
-            framework_interest_dict['agreementDetails'] = agreement_details
+        if uploader_user_id is not None:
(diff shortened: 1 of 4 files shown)


dmapiclient/data.py
```diff
@@ -181,12 +181,12 @@ def set_framework_result(self, supplier_id, framework_slug, is_on_framework, use
             user=user,
         )
 
-    def register_framework_agreement_returned(self, supplier_id, framework_slug, user, agreement_details=None):
+    def register_framework_agreement_returned(self, supplier_id, framework_slug, user, uploader_user_id=None):
```
allait: Should it be optional? As far as I understand, API requires the field to be set.

idavidmcdonald: We require older functionality for returning agreements for gcloud-7 to
still work (which does not require a uploader_user_id). The API only
requires the field to be set for frameworks with a framework agreement
version.

TheDoubleK:
:+1:

PR #29, opened 2016-07-18
Add audit event types for submitting signed framework agreement

allait:
Events created  when the user uploads a new agreement file (event created by the supplier frontend), updates agreement details (like signer details) or sets "agreement returned".


idavidmcdonald:
👍 when tests pass

PR #30, opened 2016-07-25
Specialist briefs can be open for 1 or 2 weeks.

Wynndow:
For [this ](https://www.pivotaltracker.com/story/show/123706751) story on Pivotal.

This PR adds a new flag to the find_briefs route. This is invoked by the buyer frontend when getting a list of briefs for the catalogue.

The human flag is passed to the API to make it return a list of briefs sorted in a human readable way. PR on the API can be found [here](https://github.com/alphagov/digitalmarketplace-api/pull/422)

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '5.1.0'
+__version__ = '5.2.0'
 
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
👍

PR #31, opened 2016-07-29
Remove `update_brief_status' method and add `publish_brief` method

idavidmcdonald:
- Relies on [the API pull request](url) to be merged first.

- Breaking change, bumps version to 6.0.0

- Functionality to update a brief's status to a requested value has been deprecated in the API. A `publish` route has been added to the API, which updates the brief's status from 'draft' to 'live', that should be used instead. We have therefore removed the clients `update_brief_status` method and added a `publish_brief` method to enable this transition.

- Note, whilst a `withdraw` route has also been added to the API, we are not yet adding that to the client as this route should not be used in any of our front end apps. Withdrawing functionality is not yet being given to users and instead developers will directly curl the API to withdraw an opportunity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,30 @@
 
 Records breaking changes from major version bumps
 
+## 6.0.0
+
+PR: [#31](https://github.com/alphagov/digitalmarketplace-apiclient/pull/31)
+
+### What changed
+
+Removed `update_brief_status` method.
+
+This functionality has been removed and calls to publish a brief (i.e. change it's status from
+'draft' to 'live') should now use the `publish_brief` method. There should no other current
+uses of the `update_brief_status` in our apps that set a status to something other than 'live'.
+
+### Example app change
+
+Old
+```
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,18 @@
...
+PR: [#31](https://github.com/alphagov/digitalmarketplace-apiclient/pull/31)
+
+### What changed
+
+Removed `update_brief_status` method.
+
+This functionality has been removed and calls to publish a brief (i.e. change it's status from
+'draft' to 'live') should now use the `publish_brief` method. There should no other current
+uses of the `update_brief_status` in our apps that set a status to something other than 'live'.
+
```
allait: ` ### Example app change` seems to have been lost in the past few updates, but I think we can do one for this to show how the method name and arguments should be changed.

idavidmcdonald: Thanks, have added.

allait:
Looks good 👍

PR #32, opened 2016-08-05
Add method to call /buyer/briefs on api

Wynndow:
For [this](https://www.pivotaltracker.com/story/show/124104161) story on Pivotal.

This has been created for calling from the admin frontend, to allow for downloading of a csv containing buyers info and their briefs.

PR for the new endpoint on the api is [here](https://github.com/alphagov/digitalmarketplace-api/pull/430)

PR #33, opened 2016-08-09
Add only_buyers and with_users to /users and /briefs respectively

Wynndow:
Part of [this](https://www.pivotaltracker.com/story/show/124104161) Pivotal story.

It replaces [this](https://github.com/alphagov/digitalmarketplace-apiclient/pull/32) PR from a previous version.

These updates are to allow the admin app to request briefs from the API with their user information, and to request users are filtered by `role == 'buyer'`.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -247,10 +247,15 @@ class DataAPIClient(BaseAPIClient):
                 "users": user,
             })
 
-    def find_users(self, supplier_id=None, page=None):
+    def find_users(self, supplier_id=None, page=None, role=None):
         params = {}
+        if supplier_id is not None and role is not None:
+            raise ValueError(
+                "Cannot get users by both supplier_id and role")
         if supplier_id is not None:
             params['supplier_id'] = supplier_id
+        if role is not None:
+            params['role'] = role
         if page is not None:
             params['page'] = page
         return self._get("/users", params=params)
@@ -563,7 +568,7 @@ class DataAPIClient(BaseAPIClient):
         return self._get(
             "/briefs/{}".format(brief_id))
 
-    def find_briefs(self, user_id=None, status=None, framework=None, lot=None, page=None, human=None):
+    def find_briefs(self, user_id=None, status=None, framework=None, lot=None, page=None, human=None, with_users=None):
         return self._get(
             "/briefs",
             params={"user_id": user_id,
@@ -571,7 +576,8 @@ class DataAPIClient(BaseAPIClient):
                     "lot": lot,
                     "status": status,
                     "page": page,
-                    "human": human
+                    "human": human,
+                    "with_users": with_users
                     }
         )
 
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -504,6 +504,20 @@ class TestDataApiClient(object):
 
         assert user == self.user()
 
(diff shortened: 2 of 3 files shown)


pcraig3:
Seems like you could squash your "Disallow supplier_id & only_buyers for find_users" commit into the first one.

dmapiclient/data.py
```diff
@@ -563,15 +568,16 @@ def get_brief(self, brief_id):
...
         return self._get(
             "/briefs",
             params={"user_id": user_id,
                     "framework": framework,
                     "lot": lot,
                     "status": status,
                     "page": page,
-                    "human": human
+                    "human": human,
+                    "with_users": with_users
```
pcraig3: Turns out that if we ask for a `user_id`, [then it doesn't ever return briefs with users](https://github.com/alphagov/digitalmarketplace-api/blob/b8bb54b044daf663a9a6401a344fc0cc6c451e8c/app/main/views/briefs.py#L141).  

Nothing goes wrong, just that passing both parameters results in unexpected behaviour.

Maybe throw in a warning message like you did in the `find_users` method?

pcraig3:
Also, it looks like creating branches with `/`s in them [is a weird thing to do](http://stackoverflow.com/questions/2527355/using-the-slash-character-in-git-branch-name).

dmapiclient/data.py
pcraig3: More importantly, this is what my API is getting as a request when I run this:

```
http://localhost:5000/briefs?with_users=True
```

So we're sending back `'True'`, meaning we fail the equality test in the API, (which is looking for `'true'`)[https://github.com/alphagov/digitalmarketplace-api/blob/b8bb54b044daf663a9a6401a344fc0cc6c451e8c/app/main/views/briefs.py#L119].

Wynndow: @pcraig3 Have fixed the user_id thing in the api [here](https://github.com/alphagov/digitalmarketplace-api/pull/431/files#diff-b06835e202eca80d421684ffb7ba77b2R143). Just added the with users to the serialize method.

Also, having just spoken to you about the `'True'` v `'true'` thing, I'm going to handle it in the api by lowering whatever comes through so either should work.

Wynndow:
@pcraig3 Agreed. Seems to be working for now but I'll refrain in the future!

tests/test_apiclient.py
```diff
@@ -504,6 +504,20 @@ def test_find_users_by_supplier_id(self, data_client, rmock):
...
+            data_client.find_users(supplier_id=123, role='buyer')
+
+    def test_find_users_by_role(self, data_client, rmock):
+        rmock.get(
+            "http://baseurl/users?role=buyer",
+            json=self.user(),
+            status_code=200)
+
+        user = data_client.find_users(role='buyer')
+
```
idavidmcdonald: Seems like all our other tests are checking that the mock has been called:

`assert rmock.called`

Wynndow: All the tests for finding or getting users seems to be doing it this way? Example [here](https://github.com/alphagov/digitalmarketplace-apiclient/pull/33/files#diff-6ec10ad3428f70a1a2ae819921ff6b45R537).

idavidmcdonald: Fair enough then

idavidmcdonald:
Looks good to me after a rebase 👍

PR #35, opened 2016-08-12
[#127806845]AuditTypes: add agree_framework_variation

risicle:


dmapiclient/audit.py
```diff
@@ -39,6 +39,7 @@ class AuditTypes(Enum):
     send_clarification_question = "send_clarification_question"
     send_application_question = "send_application_question"
     answer_selection_questions = "answer_selection_questions"
+    agree_framework_variation = "agree_framework_variation"
```
idavidmcdonald: Would this be better placed in the `Framework agreements` section?

idavidmcdonald: Also it's interesting that when a framework agreement is returned (which seems like just as, if not more, important than a contract variation) that we don't have a specific audit event for this. We are instead using the `supplier_update` event (as seen [here](https://github.com/alphagov/digitalmarketplace-api/blob/master/app/main/views/suppliers.py#L459). I think it seems sensible to have a more specific audit type like you are doing and maybe we should have done similar when a framework agreement is returned. @pcraig3

TheDoubleK: Yep we should definitely have had a new audit event for that `supplier_update` I think is more like changing supplier details like address or whatever. Worth a discussion if we should change it now or not.

This audit event probably does belong in the `Framework agreements` section, even though it is a thing that suppliers do.

But :+1: either way.

PR #38, opened 2016-08-22
Pluralise agreed variations

TheDoubleK:
To match changes agreed with @risicle on this API PR:
 - [ ] https://github.com/alphagov/digitalmarketplace-api/pull/435


risicle:
:+1:

PR #40, opened 2016-08-23
Preserve base URL host and scheme when requesting next API page

allait:
API responses always return next/prev links with https:// since API doesn't know if the request was made internally or through nginx. This means that apps using `_iter` methods will try to request `page=2` from `https://...`, which doesn't work when accessing the API apache server directly.

To fix this we make `_request` always use `base_url` host and scheme.


Wynndow:
👍

PR #41, opened 2016-09-02
Add agreement-related audit events

pcraig3:
We need audit events to track creating, updating, signing, and countersigning an agreement.


idavidmcdonald:
👍

PR #44, opened 2016-09-29
Add route for putting agreements on hold

Wynndow:
Part of this [story in Pivotal](https://www.pivotaltracker.com/story/show/128842457).

When a CCS admin hits the button in the admin app to put an agreement on hold, the app needs to hit the api to record this change. This route on the apiclient hits the endpoint on the api that does this, passing the current user with it for an audit event.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -669,3 +669,10 @@ class DataAPIClient(BaseAPIClient):
             },
             user=user,
         )
+
+    def put_signed_agreement_on_hold(self, framework_agreement_id, user):
+        return self._post_with_updated_by(
+            "/agreements/{}/on-hold".format(framework_agreement_id),
+            data={},
+            user=user
+        )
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -248,6 +248,13 @@ def find_framework_suppliers(self, framework_slug, agreement_returned=None):
             params=params
         )
 
+    def put_signed_agreement_on_hold(self, framework_agreement_id, user):
```
idavidmcdonald: Would have been hard to spot but we have started a new section for framework agreement routes towards the bottom of the file (around line 650). This should be grouped with those please

idavidmcdonald:
Nice one. After the short comment about location of the function has been addressed then this will be all good to go 👍

dmapiclient/data.py
Wynndow: Nice. I did look but before I rebased so completely missed it!

PR #45, opened 2016-09-30
Remove temporary route to set countersigned timestamp

TheDoubleK:
This route only existed for a script. Now it's been run so we can remove it.


idavidmcdonald:
👍

PR #47, opened 2016-10-04
Add countersign agreement route

Wynndow:
Part of (this story on Pivotal)[https://www.pivotaltracker.com/story/show/128842457].

The admin frontend needs to be able to mark `framework_agreements` as countersigned. There's a new route on the API (PR is (here)[https://github.com/alphagov/digitalmarketplace-api/pull/466]) to deal with this.

This route on the apiclient hooks the two together.

It may need updating in future depending on how the `countersigned_agreement_path` is going to be set.

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -1206,24 +1206,26 @@ class TestDataApiClient(object):
     def test_put_signed_agreement_on_hold(self, data_client, rmock):
         rmock.post(
             "http://baseurl/agreements/101/on-hold",
-            json={},
+            json={'David made me put data in': True},
             status_code=200
         )
 
-        data_client.put_signed_agreement_on_hold(101, 'Chris')
+        result = data_client.put_signed_agreement_on_hold(101, 'Chris')
 
+        assert result == {'David made me put data in': True}
         assert rmock.call_count == 1
         assert rmock.last_request.json() == {'updated_by': 'Chris'}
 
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -1215,6 +1215,18 @@ def test_put_signed_agreement_on_hold(self, data_client, rmock):
         assert rmock.call_count == 1
         assert rmock.last_request.json() == {'updated_by': 'Chris'}
 
+    def test_countersign_agreement(self, data_client, rmock):
+        rmock.post(
+            "http://baseurl/agreements/101/countersign",
+            json={},
```
Wynndow: Yeah, I had a think about that. As far as I can tell all testing the result would do would be testing the mock I created. The below is a passing test for example.

```
def test_countersign_agreement(self, data_client, rmock):
        rmock.post(
            "http://baseurl/agreements/101/countersign",
            json={'socks': True},
            status_code=200
        )

        result = data_client.countersign_agreement(101, 'Chris')

        assert result == {'socks': True}
        assert rmock.call_count == 1
        assert rmock.last_request.json() == {'updated_by': 'Chris'}
```

I'm not sure why the other tests are testing that way. What do you think?

Wynndow: Ah, yep, it's testing that something is being returned.

idavidmcdonald:
👍 Good work. Will be good to good after the API pull request has been merged in.

PR #49, opened 2016-10-12
Add 'agreement' key to expected json data to approve an agreement

idavidmcdonald:
All API calls should have a top level key that represents the object that is being changed. We had forgotten the 'agreement' key and had been sending the framework agreement data in the top level so this has been fixed to be consistent with the rest of our app.

See related [API pull request](https://github.com/alphagov/digitalmarketplace-api/pull/472).


TheDoubleK:
:+1:

PR #50, opened 2016-10-13
Filter agreements by status

TheDoubleK:
~~Forthcoming~~ [This pull request on the API](https://github.com/alphagov/digitalmarketplace-api/pull/474) adds filtering by status to the list framework agreements endpoint.  

This adds functionality to the API client to be able to pass a status parameter in.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -245,7 +245,7 @@ class DataAPIClient(BaseAPIClient):
                                    returned their framework agreement. If 'agreement_returned' is set then
                                    any value for 'statuses' will be ignored.
         :param statuses: A comma-separated list of the statuses of framework agreements that should be returned.
-                         Valid statuses are: draft, signed, on-hold, approved and countersigned.
+                         Valid statuses are: signed, on-hold, approved and countersigned.
         '''
         params = {}
         if agreement_returned is not None:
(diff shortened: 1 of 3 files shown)


TheDoubleK:
Rebased and version umber bumped further.

idavidmcdonald:
👍 (would be good to wait for the API to get a thumbs up before merging this just in case...)

dmapiclient/data.py
```diff
@@ -239,10 +239,19 @@ def agree_framework_variation(self, supplier_id, framework_slug, variation_slug,
...
+                                   returned their framework agreement. If 'agreement_returned' is set then
+                                   any value for 'statuses' will be ignored.
+        :param statuses: A comma-separated list of the statuses of framework agreements that should be returned.
+                         Valid statuses are: signed, on-hold, approved and countersigned.
+        '''
         params = {}
         if agreement_returned is not None:
             params['agreement_returned'] = bool(agreement_returned)
+        elif statuses is not None:
+            params['status'] = statuses
```
risicle: I'd consider this `elif` to be inserting too much business-logic into the apiclient. It knows too much about the nature of these `SupplierFrameworks` - that you won't get any valid statuses if you're not setting `agreement_returned`.

In fact, I'd argue that the behaviour is wrong - you should be able to filter by `status` without filtering by `agreement_returned`. The latter would just be implicit.

PR #54, opened 2016-10-27
Create submit brief response route

Wynndow:
Part of this story on Pivotal [https://www.pivotaltracker.com/story/show/129838783](https://www.pivotaltracker.com/story/show/129838783).

There's a new view on the API to allow brief_responses to be submitted. This route allows frontend apps to call it.


idavidmcdonald:
👍 good when tests green

PR #55, opened 2016-10-27
Send back page questions when creating a brief response

idavidmcdonald:
Similar to how is done for creating a brief.

We will therefore be able to validate certain questions when creating a brief response (rather than having to validate all of them or not validating any of them).

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -615,12 +615,13 @@ class DataAPIClient(BaseAPIClient):
             params={"supplier_id": supplier_id}
         )['services']) > 0
 
-    def create_brief_response(self, brief_id, supplier_id, data, user):
+    def create_brief_response(self, brief_id, supplier_id, data, user, page_questions=None):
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -615,12 +615,13 @@ def is_supplier_eligible_for_brief(self, supplier_id, brief_id):
             params={"supplier_id": supplier_id}
         )['services']) > 0
 
-    def create_brief_response(self, brief_id, supplier_id, data, user):
+    def create_brief_response(self, brief_id, supplier_id, data, user, page_questions=None):
```
idavidmcdonald: I guess so. I've looked at how we are doing this for the other API client methods and found at 3 methods that do it this way so I'd be tempted to keep with convention (to be honest I don't see a particular benefit/disadvantage or doing it either way).

idavidmcdonald:
👍 self thumbs up as Tom is interviewing but says is good to go after his comment was addressed

PR #57, opened 2016-11-15
Update brief response route

Wynndow:
Part of this story on Pivotal [https://www.pivotaltracker.com/story/show/129843611](https://www.pivotaltracker.com/story/show/129843611)

The new flow for brief_responses on the supplier frontend needs to be able to send updates to brief_responses to the api. This adds a route on the client to allow it.


idavidmcdonald:
👍 nice one

PR #58, opened 2016-11-24
Throwing if particular API client is not configured.

galund:
Fix crash on status URI when some API is not actually configured in the environment.

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -8,11 +8,14 @@ except ImportError:
 
 import requests
 from flask import has_request_context, request, current_app
+
 import backoff
 from monotonic import monotonic
 
 from . import __version__
 from .errors import APIError, HTTPError, HTTPTemporaryError, InvalidResponse
+from .exceptions import ImproperlyConfigured
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


dmapiclient/base.py
```diff
@@ -76,6 +76,9 @@ def _delete_with_updated_by(self, url, data, user):
     def _request(self, method, url, data=None, params=None):
         if not self.enabled:
             return None
+        if self.base_url is None:
+            logger.info("{} has no URL configured".format(self.__class__.__name__))
```
allait: I'm not sure about this. This is an error condition in any actual environment, so why would we want to hide this?

allait: This should never work like this in a production app on PaaS, so I don't think it makes sense to merge this into master since it hides a potential problem from monitoring and health checks. Adding search API to the PaaS environment is what we'd do when setting up an actual environment.

In the meantime, since we're running a throwaway prototype branch on PaaS, so I'm happy with any hack on that branch to fix issues caused by this, but I suggest we close this PR.

galund: Okay, in which case for master can we make it raise instead of log? Just because the error message that results from a missing API config isn't very clear at the moment.

galund: Hmm, so the reason that the constructor can't be the thing that fails, is because in practice the constructor is always called with no arguments (e.g. api:app/__init__.py:13), and the base URL is set later (via an init_app call).

Not sure why we have both ways of setting the configuration, but there it is.

galund:
OK this seems right to me now

allait:
👍

PR #59, opened 2016-12-30
init_app: annotate self onto app instance provided in both DataAPIClient & SearchAPIClient

risicle:
Following a standard pattern from flask extensions, I've made `DataAPIClient` and `SearchAPIClient`'s `init_app` methods annotate themselves onto the provided `app` instance via `data_api_client` and `search_api_client` attributes respectively.

This would allow frontend code to access the api clients in a more uniform & robust way (through `current_app`) than they are currently doing. I have a branch of the admin frontend with associated changes brewing as a proposal/strawman - this PR is intended to go with that - on its own this PR is meaningless, so just disregard until then.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '7.11.0'
+__version__ = '7.12.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
 from .errors import REQUEST_ERROR_STATUS_CODE, REQUEST_ERROR_MESSAGE  # noqa
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -9,6 +9,8 @@ class DataAPIClient(BaseAPIClient):
         self.base_url = app.config['DM_DATA_API_URL']
         self.auth_token = app.config['DM_DATA_API_AUTH_TOKEN']
 
+        app.data_api_client = self
+
     # Audit Events
 
     def find_audit_events(
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -14,6 +14,8 @@ class SearchAPIClient(BaseAPIClient):
         self.auth_token = app.config['DM_SEARCH_API_AUTH_TOKEN']
         self.enabled = app.config['ES_ENABLED']
 
+        app.search_api_client = self
+
     def _url(self, index, path):
         return u"/{}/services/{}".format(index, path)
 
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -196,6 +196,7 @@ class TestSearchApiClient(object):
         assert search_client.base_url == "http://example"
         assert search_client.auth_token == "example-token"
         assert not search_client.enabled
+        assert app.search_api_client is search_client
 
     def test_get_status(self, search_client, rmock):
         rmock.get(
@@ -385,6 +386,7 @@ class TestDataApiClient(object):
 
         assert data_client.base_url == "http://example"
         assert data_client.auth_token == "example-token"
+        assert app.data_api_client is data_client
 
     def test_get_status(self, data_client, rmock):
         rmock.get(


allait:
I haven't seen any of the extensions do this TBH. I think the global module variable is the approach all of the extensions we're currently using are taking and it's also the approach explained in the [Flask documentation on extensions](http://flask.pocoo.org/docs/0.12/extensiondev/#the-extension-code).

Is this solving an existing problem, because even setting the argument about common practice aside, I'm not really sure what the benefits are of this.

risicle:
https://github.com/maxcountryman/flask-login/blob/master/flask_login/login_manager.py#L113

If the "standard flask way" of putting these objects in as global variables than it's just another dumb piece of flask design as far as I'm concerned. It makes it near-impossible to reliably mock-patch (not that half of our tests don't *try* to, bless 'em) because most of our view modules do `from app import data_api_client`, meaning that it's not globally mockable, only module-local mockable. Which means that tests go and module-locally mock-patch `data_api_client`, so surprise surprise, if you ever try and use a `data_api_client` from a different module - the tests get an unmocked `data_api_client`. If you want I'll show you the example changes I made to the admin frontend to work with this and how much neater it makes things.

allait:
I'm not sure I agree with your definition of "reliably mock-patch". It requires you to know which module you're patching and doesn't apply a blanket patch on all references if you're trying to replace the entire client instance. It still allows you to patch methods globally, which I think we do for common things like auth.

The "new reference on `from import`" can be an annoying issue sometimes, but I don't believe it is in this case. Most of the uses of API client are inside view modules, which shouldn't really call each other directly. Things like helpers should generally accept `apiclient` instances as an argument instead of trying to import.

risicle:
I can't really believe I'm having this conversation. Arguing that flawed infrastructure limitations are a good thing - I just can't really.. if you're going to stop people using the `data_api_client` outside view modules at least make it explicit, not just something which works fine in real life but you get mysterious errors when you try and run your tests against them.

Even so, it is *not at all* a good restriction to make, more another thing to cripple the ability for our code to evolve into something more well-factored.

danielknell:
from experience, most extensions store a reference on the application object, either directly or as a key in a dict property called `extensions`, and these references are used by all the extensions internals.

storing a central reference and then perhaps swapping out the global for a proxy object could have open some interesting potential futures, we could perhaps have a smarter mock that can have request-response pairs registered with it, or be able to optionally wrap the instance on dev for enhanced debugging without taking that baggage to prod to name a couple.

if the globals are our way forward maybe we could add some custom pyflakes rules to ensure that folk are doing dependency injection and not relying on the global outside of view functions? would help keep us all honest.

allait:
That's the thing, I haven't seen any actual examples of things `current_app` attribute allows us to do that's impossible or substantially more difficult with the module globals. And switching all apps from module globals and their mocks to `current_app` seems like a lot of work, so I feel that the benefits have to be significant to make it worthwhile. 

I also disagree with the "flawed infrastructure limitations" assessment. You can't call one thing "good" and another thing "bad" without explanation and expect people to take this as an honest argument. I have a strong feeling that the proposed approach is more prone to hiding potential issues behind an "easy to globally mock everything" interface.

(As a note, if we're talking about "how would this be better in theory" – I'd get rid of the global instances altogether and instantiate clients on demand. They're stateless so there's not much point in sharing them and mocking individual methods on a class seems like a more robust approach overall.)

risicle:
My argument is actually to not use mock-patching at all for `data_api_client`, I was just pointing out that the current way of doing it is flawed (mostly because it encourages people to mock it in tests in an incomplete way and gives them little alternative except to write many lines of places it *also* should patch in every single patch request). The way I was going to suggest test mocking for `data_api_client` is in another PR which isn't fully prepared yet, but you jumped on this one before I could get that ready.

In short, I think the reason the current way of doing it is it suggests two different mutually incompatible way of doing things. _Either_ you do things strictly through dependency injection (which imho gets more than a bit ugly) _or_ you do things through globals. The current way dangles a global in front of you but then punishes you with weird errors when you try to write a working test suite for your code if you use it.

risicle:
https://github.com/alphagov/digitalmarketplace-admin-frontend/compare/ris-old-test-fw-purge...ris-app-apiclient-refactor#diff-560ef0f937b6c64463419b8a77b1e670

Unfortunately one branch sits on top of another (necessarily) so it's a `compare` between two of my branches

allait:
@risicle I honestly don't how that change makes a big difference.

It avoids the use of `mock.patch` (which I don't think we have a problem with) by adding a function argument that is only useful for testing. It makes it harder to tell which tests are using the API client and therefore need to mock it and which aren't (because it's now always mocked). And it ties everything to `current_app` which I personally find a bit awkward to use in tests because it's hard to tell whether what you're about to do requires application context or not (and multiple redundant `with app.app_context()` we have in our tests feel like proof that I'm not alone in this).

I think the biggest difference is that it replaces the explicit "I'm using apiclient for this view and that's why I need to mock this" with an implicit "tests run with a special app that has a fake apiclient", and I'm not sure I see the benefits of this.

risicle:
I think the latter is *much* better. I don't see how the current situation of randomly "ha-*ha* guess why!" breaking in the test phase (and only the test phase) if you don't follow the unwritten rules is at all tenable. If you *really* want it to explicitly only work if you've explicitly said this test should have a mocked api we could use the pytest mark mechanism to make this only work in marked tests (in unmarked tests it could be replaced with a `None` so you don't a) have a weird inexplicable error if you've forgotten to add the mark and b) risk having tests running around with a semi-configured api client.

I don't know how to even reason with you if you really *truly* think the current way we have things is better than any of these alternatives. I don't understand this "worse is better" argument. It's not. It's worse. If you don't think modules should use it as a global variable, get rid of it as a global variable.

allait:
> breaking in the test phase (and only the test phase) if you don't follow the unwritten rules

How is the new approach any different? You either set up your test environment right and your test is working or you don't. The only difference is the assumptions the default environment makes and the errors that those assumptions can hide.

With `mock.patch` the assumption is that the view doesn't make any API requests, and the error happens when it does (at which point you need to set up the mock). With `current_app` the assumption is that the view is using API client and the hidden error is that it actually IS making requests when you didn't expect it to, including requests that you didn't set up proper return values for, so I highly doubt that error messages you get in those cases are any better than the ones you get when you forget to mock now.

> risk having tests running around with a semi-configured api client.

Tests are running in a test environment and on Travis, which doesn't have any other apps running, so I'm not sure what the risk is. If the problem is that we don't get a useful error message when a request isn't mocked, then that's something worth investigating. I don't think this actually solves it though. It replaces the "API client actually tried to send the request and failed" error with "mock API client returned a generic mock object instead of JSON response" error when you forget to configure the mock, which I'm not at all convinced is better.

> I don't understand this "worse is better" argument. It's not. It's worse.

I don't think I've made the "worse is better" argument here. But if I did, I'd say that we have a lot of prioritised things that we have to be doing and I feel like this is not the best way to solve any of them, help with any of them or make any of them easier.

danielknell:
> And it ties everything to current_app which I personally find a bit awkward to use in tests because it's hard to tell whether what you're about to do requires application context or not (and multiple redundant with app.app_context() we have in our tests feel like proof that I'm not alone in this).

you can resolve the app context stuff with some py.test fixture work, either ensuring all tests run inside an app context via an `autouse` fixture or specific tests by `mark.usefeature`.

example: https://github.com/artisanofcode/flask-generic-views/blob/master/tests/conftest.py#L12

risicle:
> How is the new approach any different?

If you *really* wanted to have to explicitly mark whether tests are requiring the api client or not (which I'm not at all convinced is worthwhile) it would be possible to use the presence of a pytest mark to switch the mock api client out for `None` or similar if the mark is not present.

> Tests are running in a test environment and on Travis

Well they're also running on our local machines which are regularly configured against local databases, occasionally the preview db and occasionally the staging db and developers sometimes set settings on the wrong environment in `config.py` (I know I have several times).

> But if I did, I'd say that we have a lot of prioritised things that we have to be doing and I feel like this is not the best way to solve any of them, help with any of them or make any of them easier.

This is a spurious argument, It's not about prioritization - the work is largely already done and I'd be willing to do remaining work in my own time to fix ugly corners of the code like this.

But still - you haven't answered - if we're preferring dependency injection, why do we have these misleading globals around at all? Removing these globals would also allow us to get rid of the really ugly backwards-relative imports a la

```
from ... import data_api_client
```

If it comes down to preferring dependency injection, the reason *I* would _prefer_ we avoid that approach is I think it would be better to model our use of the api client on the way flask uses database connections. You don't have flask apps needing to pass database connections around in function args...

I've also not been able to find any of the flask docs or tutorials which show the use of a global like this that isn't an example for a single file project. Flask extensions add themselves to the app (or more commonly `app.extensions[]`) for a reason.

allait:
> it would be possible to use the presence of a pytest mark

I think a lot of the arguments so far have been about how we could implement things to fix the issues the new approach creates. While I don't doubt it's possible, it's additional work on top of the work we don't have to do in the first place. We could also use all that effort to fix the existing issues with the current approach instead.

> regularly configured against local databases, occasionally the preview db and occasionally the staging db and developers sometimes set settings on the wrong environment in config.py

If this is the actual problem you're worried about, a much easier solution would be to swap the URLs in test config to not point to the local services. This change does little for other potential errors in `config.py`, since the can affect running apps instead of test environments.

> This is a spurious argument, It's not about prioritization - the work is largely already done

It absolutely is about prioritization. I do not agree that the work is done since I think moving only one app out of 3 creates a much bigger problem on it's own. I'm also questioning the decision to start working on this to begin with, which feels to me like a prioritization issue as well. And the work is not done until things are running in production, which means all of this needs to be reviewed, tested and deployed, so I think you're underestimating the efforts of other people on the team that would have to be involved in this one way or the other.

> if we're preferring dependency injection, why do we have these misleading globals around at all?

Again, I'm not really sure what the issue is. The globals are convenient to use in the view functions. I'd much rather the helper function accepted API client instances as arguments, since this avoids the need to patch the method in more than one place. If you look at your own admin app change – each test has only one patch to the view API client that you're replacing, so it feels to me that's exactly how things are working already.

> model our use of the api client on the way flask uses database connections. [...] I've also not been able to find any of the flask docs or tutorials which show the use of a global like this that isn't an example for a single file project.

Database connection is a module global in the API, so I'm not really sure how this is an argument for the app reference.

risicle:
> I think a lot of the arguments so far have been about how we could implement things to fix the issues the new approach creates.

That's only because you have been steering it in that direction largely with arguments that make no sense. This feels a lot like a smokescreen.

> I'm also questioning the decision to start working on this to begin with, which feels to me like a prioritization issue as well

Paraphrased quote from meeting (3 weeks ago?) "I don't know why we're having this conversation - you don't need to ask to work on this kind of stuff, you just do it...". It was a single afternoon when I was blocked on other stories that I used to create an example of a proposed change. Saying this sort of thing is not allowed is akin to saying that large-scale changes are not allowed by anybody who isn't you.

> Database connection is a module global in the API, so I'm not really sure how this is an argument for the app reference.

Right, and you would never try to mock out the database connection through `mock.patch`, would you? Because you know it wouldn't actually catch all the references and uses.

It's also not meant to be an argument _for_ the _app reference_ - please re-read the sentence if you're unsure - it's an argument _against_ the dependency injection approach, though not one I feel so strongly about. But I think you know that.

danielknell:
Does adding the reference to `current_app` actually break anything in how the api clients currently functions?

risicle:
So pray tell how are we supposed to mock out `data_api_client` for tests which perform a `follow_redirects=True` when it redirects between two view modules? Mock each view module's `data_api_client`s separately and juggle the two mocks?

risicle:
This is turning into a meeting next week about the broader future of testing/mocking/data generation for the frontends next week sometime.

PR #60, opened 2017-01-05
adding publishedAt to the brief stub when status is closed

danielknell:
closed briefs have a publishedAt date too and we are using that to decide which flow to use for things now, this should make those tests easier and fix existing tests that run over old flows.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '7.11.0'
+__version__ = '7.12.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


idavidmcdonald:
With your suggestion though @allait, you would be able to have a closed brief with clarification questions being `True`. Is that valid?

allait:
@idavidmcdonald that would never happen in an actual API response, but neither would the missing key, so I'd say it's still a step in the right direction. We can force the flag based on the status to avoid potential errors, but I'm not sure if it's a big issue.

danielknell:
@allait is that better?

allait:
Yeah, looks good 👍

PR #61, opened 2017-01-24
[#134795163]get_supplier_frameworks: add with_declarations support

risicle:
Story https://www.pivotaltracker.com/story/show/134795163

kwarg defaulting to `None` to which omits url parameter (uses server's default).

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -1220,6 +1220,17 @@ class TestDataApiClient(object):
         assert result == {'supplierFrameworks': [{"agreementReturned": False}, {"agreementReturned": True}]}
         assert rmock.called
 
+    def test_find_framework_suppliers_no_declarations(self, data_client, rmock):
+        rmock.get(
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -1220,6 +1220,17 @@ def test_find_framework_suppliers_with_statuses(self, data_client, rmock):
         assert result == {'supplierFrameworks': [{"agreementReturned": False}, {"agreementReturned": True}]}
         assert rmock.called
 
+    def test_find_framework_suppliers_no_declarations(self, data_client, rmock):
+        rmock.get(
+            'http://baseurl/frameworks/g-cloud-7/suppliers?with_declarations=false&',
```
TheDoubleK: Why do we get the trailing `&` here? Tests for e.g. filtering by status don't get a trailing & on the url.

risicle: Brain fart.

PR #62, opened 2017-01-27
Kill the import_* wrapper methods

pcraig3:
Now that we have a better way of doing data migrations onto local dev machines, we can obliterate the supplier/service import routes.

The only place these methods are used is in the import script found in the API. A subsequent pull request on the API will remove the script as well as the routes that actually do the imports.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -101,12 +101,6 @@ class DataAPIClient(BaseAPIClient):
             "/suppliers/{}".format(supplier_id)
         )
 
-    def import_supplier(self, supplier_id, supplier):
-        return self._put(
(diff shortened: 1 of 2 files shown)


TheDoubleK:
You need moar test coverage, dawg.

TheDoubleK:
These are also used by the functional tests, and @idavidmcdonald is just adding some new stuff that uses at least the `import_service` route.

Ah, but maybe we don't need the client methods for that though? Seeing as it's all Ruby.

TheDoubleK:
:+1:

idavidmcdonald:
Yep, the 'current' functional tests use the import route on the API although call it directly as it's ruby.
I'm looking at the new functional tests and these may also need to use that route too...

PR #63, opened 2017-01-31
Bump version after removing import routes

pcraig3:
- Updated the changelog as technically this is a breaking change
- Fixed a tiny link text issue
- Added syntax highlighting to code samples in the changelog


pcraig3:
🎱

PR #67, opened 2017-02-10
138323285-rm_find_frameworks_iter__add_find_supplier_declarations

benvand:
* remove `find_frameworks_iter` (never used, broken)
* add `find_supplier_declarations`
* version bump
* test

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -140,6 +140,11 @@ class DataAPIClient(BaseAPIClient):
             user=user,
         )
 
+    def find_supplier_declarations(self, supplier_id):
+        return self._get(
+            "/suppliers/{}/frameworks".format(supplier_id)
+        )
+
     def get_supplier_declaration(self, supplier_id, framework_slug):
         response = self._get(
             "/suppliers/{}/frameworks/{}".format(supplier_id, framework_slug)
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -140,6 +140,12 @@ def register_framework_interest(self, supplier_id, framework_slug, user):
             user=user,
         )
 
+    def find_supplier_declarations(self, supplier_id):
+        response = self._get(
+            "/suppliers/{}/frameworks".format(supplier_id)
+        )
+        return {'declarations': response['frameworkInterest']}
```
TheDoubleK: We don't generally have the ApiClient muddy the API response at all - we just pass the response as-is. Looks like the only time we do wrap things I a new key is in `get_supplier_declaration`, which I suspect is where you got the idea to do the same thing here. I don't think that's a practice we want to spread though, and this method would be better just passing the `frameworkInterest` response through unmodified.

TheDoubleK: That should have been "modify" rather than muddy... on my stupid phone :)

benvand: So I've changed it to just return `response`
Is your suggestion that returning `response['frameworkInterest']` is a better idea?

TheDoubleK: No just the passing the whole response is the right thing I think.

TheDoubleK:
:+1:

PR #69, opened 2017-02-17
Fix enum34 install error on Python 3.6

allait:
enum34 1.1.2 failed to install on Python 3.6, so I'm bumping it to
the latest version (1.1.6) which has a pre-built wheel.


TheDoubleK:
:+1:

PR #77, opened 2017-06-19
Add earliest_for_each_object to audit event

MuriloDalRi:
Update APIClient according to [#596](https://github.com/alphagov/digitalmarketplace-api/pull/596)

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '8.12.0'
+__version__ = '8.13.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


risicle:
Ah - version bump

MuriloDalRi:
Better now?

risicle:
yup, but we'd better hold off on merging it till the api implementation is merged

dmapiclient/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '8.10.0'
+__version__ = '8.20.0'
```
samuelhwilliams: Should be 8.11.0

TheDoubleK: Nice spot, although we might be up to 20 by the time this all lands :p

PR #79, opened 2017-06-28
Try again when getting back 502 errors

pcraig3:
Previously, it was only 503 or 504 errors would cause the `find_{model}_iter` function to retry the failing call. Now we've added 502 errors as well.

This is to mitigate a problem we're experiencing with exporting brief responses.  It's happened a few times that the `find_brief_responses_iter` function gets through about ~190 pages of brief responses and then fails upon getting back a 502 error.

Manually testing getting back brief responses fails intermittently but not reliably so we're going to make the API client more resilient while we're investigating the problem.

This is a short-term-type solution so that we can complete the data exports, but it doesn't actually do anything to address the failures in the API.

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -2217,11 +2217,14 @@ class TestDataAPIClientIterMethods(object):
 
         assert len(results) == 2
 
+    @pytest.mark.parametrize("status_code", [
+        502, 503, 504
+    ])
     @mock.patch('time.sleep')
-    def test_find_services_backoff_on_503(self, sleep, data_client, rmock):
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -2217,11 +2217,14 @@ def test_find_services_iter_additional_arguments(self, data_client, rmock):
 
         assert len(results) == 2
 
+    @pytest.mark.parametrize("status_code", [
+        (502), (503), (504)
```
allait: Parentheses around numbers don't do anything in this case, it's safe to remove them

PR #81, opened 2017-07-11
Add 'with_clarification_questions' parameter to `find_briefs`

CrystalPea:
This is part of a story to tell interested suppliers that a question has been answered.

Related dm-api pull request: https://github.com/alphagov/digitalmarketplace-api/pull/614

Trello ticket: https://trello.com/c/2noI0EFB/472-tell-interested-suppliers-that-a-question-has-been-answered

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
 *.egg-info/
 *.pyc
-venv
+venv*
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.egg-info/
 *.pyc
+venv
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 *.egg-info/
 *.pyc
+venv
```
benvand: I name my virtualenvs `venv` and `venv3` respectively. Although this might not be a problem for much longer with the imminent move to py3 can we have the `venv*` pattern here like in https://github.com/alphagov/digitalmarketplace-api/blob/master/.gitignore ?

CrystalPea: Ok, I changed this now, thanks for spotting it :)

PR #84, opened 2017-08-15
Direct Award Projects

samuelhwilliams:
## Summary
Add APIClient methods for interacting with Direct Award projects on the API. Adds methods on the Search API client to generate the full search-api URL that a given call will hit. This is stored by the API for Project searches.

This PR also wraps the `register_uri` function of the rmock and inserts the `complete_qs` parameter, if not supplied. This forces the URLs to match exactly; without it, missing query params would not trigger a failed test case. A couple of tests broke from this change, so fixed those up to properly mock the full URL.

## Other PRs:
API: https://github.com/alphagov/digitalmarketplace-api/pull/624
Buyer FE: https://github.com/alphagov/digitalmarketplace-buyer-frontend/pull/581 (for trying out the flow only; not for review yet)

## Ticket
https://trello.com/c/pbJiKaVl/625-api-for-direct-award-projects

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -75,10 +75,7 @@ class BaseAPIClient(object):
         data = dict(data, updated_by=user)
         return self._delete(url, data)
 
-    def _request(self, method, url, data=None, params=None):
-        if not self.enabled:
-            return None
-
+    def _build_url(self, url, params):
         if not self.base_url:
             raise ImproperlyConfigured("{} has no URL configured".format(self.__class__.__name__))
 
@@ -90,6 +87,17 @@ class BaseAPIClient(object):
         url = urlparse.urlparse(url)
         base_url = urlparse.urlparse(self.base_url)
         url = url._replace(netloc=base_url.netloc, scheme=base_url.scheme).geturl()
+
+        r = requests.PreparedRequest()
+        r.prepare_url(url=url, params=params)
+
+        return r.url
+
+    def _request(self, method, url, data=None, params=None):
+        if not self.enabled:
+            return None
+
+        url = self._build_url(url, params)
 
         logger.debug("API request {method} {url}",
                      extra={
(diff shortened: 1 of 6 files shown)


dmapiclient/data.py
```diff
@@ -764,3 +764,68 @@ def unapprove_agreement_for_countersignature(self, framework_agreement_id, user,
...
+    def get_project_search(self, user_id, project_id, search_id):
+        return self._get(
+            "/projects/{}/searches/{}".format(project_id, search_id),
+            params={
+                "user-id": user_id
+            }
+        )
+
+    def find_project_services(self, user_id, project_id):
+        pass
```
galund: pedantry, but throw `NotImplementedError` here

samuelhwilliams: Ah good idea, yes.

dmapiclient/base.py
```diff
@@ -91,6 +88,17 @@ def _request(self, method, url, data=None, params=None):
         base_url = urlparse.urlparse(self.base_url)
         url = url._replace(netloc=base_url.netloc, scheme=base_url.scheme).geturl()
 
+        r = requests.PreparedRequest()
+        r.prepare_url(url=url, params=params)
```
samuelhwilliams: A direct request doesn't give us access to the full URL without sending the request. For Direct Award projects we need to generate the full Search API URL to store it in our database but don't need to actually send the request, so we are using the request lib's intermediate step (PreparedRequest) to build the URL and expose it to the other part of our application as required.

katstevens: Aha - makes sense, thanks.

dmapiclient/data.py
```diff
@@ -764,3 +764,67 @@ def unapprove_agreement_for_countersignature(self, framework_agreement_id, user,
...
+    def get_direct_award_project_search(self, user_id, project_id, search_id):
+        return self._get(
+            "/direct-award/projects/{}/searches/{}".format(project_id, search_id),
+            params={
+                "user_id": user_id
+            }
+        )
+
+    def find_direct_award_project_services(self, user_id, project_id):
+        raise NotImplementedError()
```
samuelhwilliams: There isn't an equivalent `create_` and this is basically our only `get_` method for it (as you can't get services without both of these things). Maybe this would be better of just called `get_`, but for some reason the majority of apiclient methods start with `find_` instead.

PR #86, opened 2017-08-22
Remove default index from searchapiclient

samuelhwilliams:
## Summary
We are removing the default index name from the Search API Client and requiring it to be passed explicitly. This is because we are moving away from a single index called 'g-cloud' and moving towards an index-per-framework where the index name is the framework slug. Given this, we can get the index name programmatically and don't need to statically define it anywhere.

Bump version to 9.0.0

## Ticket
Required work under https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,24 @@
 
 Records breaking changes from major version bumps
 
+## 9.0.0
+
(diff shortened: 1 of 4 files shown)


samuelhwilliams:
Ugh, coveralls!

PR #87, opened 2017-08-22
update apiclient for updated api parameter formats

samuelhwilliams:
## Summary
Update api client to use adjusted query params, JSON formats, etc for Direct Award based on feedback from code review. Query params = snake case, JSON data = camel case.
Bump version to 9.1.1


galund:
ndencies... so.. circular... depe...

PR #88, opened 2017-08-24
Save a search and create a project

MuriloDalRi:
## Summary

* Adding a new method to turn the search api url into frontend url
* Bump version to 9.2.0
* Also found a bug with two previously implemented methods for getting search/aggregation URLs whereby it would include the `q` parameter by default if not supplied. Made sure this doesn't happen and pinned it with two tests.

## Ticket
https://trello.com/c/Hlq5mzAy/619-save-a-search-and-create-a-project

### DIFF
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -52,7 +52,7 @@ class SearchAPIClient(BaseAPIClient):
 
         return self._build_url(url=self._url(index=index, path=path), params=params)
 
-    def deconstruct_url(self, search_api_url):
+    def get_frontend_params_from_search_api_url(self, search_api_url):
         """
         Converts a searchAPI url to url params a frontend understands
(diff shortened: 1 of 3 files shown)


dmapiclient/search.py
```diff
@@ -23,10 +48,22 @@ def get_url(self, path, index, q, page=None, aggregations=[], **filters):
         elif page:
             params['page'] = page
 
-        self._add_filters_to_params(params, filters)
+        self._add_filters_prefix_to_params(params, filters)
 
         return self._build_url(url=self._url(index=index, path=path), params=params)
 
+    def deconstruct_url(self, search_api_url):
```
samuelhwilliams: Just looking at this again, we could probably name this slightly better.

MuriloDalRi: We could!

PR #90, opened 2017-08-31
Remove unused param from get_direct_award_project

samuelhwilliams:
## Summary
`user-id` was being passed to the `get_direct_award_project` endpoint, but the API just ignores this. So let's remove it. And bump version to 10.0.0


TheDoubleK:
:+1:

PR #91, opened 2017-09-04
Get index from search api url

samuelhwilliams:
## Summary
To display the search overview page correctly we need to know what framework instance the search relates do. We know this from the search API URL which includes the framework slug. Therefore, we need to be able to extract the slug from the API URL. This adds a method on the search API client to do this.

Bumps version to 10.2.0.

## Ticket
https://trello.com/c/9YtT8DmX/675-task-list-saved-search-overview-page

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -389,6 +389,17 @@ class TestSearchApiClient(object):
     def test_get_aggregations_url(self, search_client):
         assert search_client.get_aggregations_url('g-cloud-9') == 'http://baseurl/g-cloud-9/services/aggregations'
 
+    @pytest.mark.parametrize('search_api_url, expected_index',
+                             (
+                                 ('http://localhost/g-cloud-8/services/search', 'g-cloud-8'),
+                                 ('http://localhost/g-cloud-9/services/search', 'g-cloud-9'),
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -389,6 +389,16 @@ def test_get_search_url(self, search_client):
     def test_get_aggregations_url(self, search_client):
         assert search_client.get_aggregations_url('g-cloud-9') == 'http://baseurl/g-cloud-9/services/aggregations'
 
+    @pytest.mark.parametrize('search_api_url, expected_index',
+                             (
+                                 ('http://localhost/g-cloud-8/services/search', 'g-cloud-8'),
+                                 ('http://localhost/g-cloud-9/services/search', 'g-cloud-9'),
+                                 ('https://search-api.preview.marketplace.team/g-cloud-9/services/search', 'g-cloud-9'),
+                                 ('https://search-api.preview.marketplace.team/g-cloud-8/services/search', 'g-cloud-8')
+                             ))
```
katstevens: Is it worth adding a case for a non-matching URL path that returns `None`?

samuelhwilliams: Yes, it is. Thanks.

PR #92, opened 2017-09-06
Lock projects and get all services by Search API URL

samuelhwilliams:
## Summary
Adds a new method to the DataAPIClient to lock a Direct Award Project, and a new method (+ `_iter`) to the SearchAPIClient to retrieve services from a given Search API URL.
Bumps the version to 10.3.0

## Ticket
https://trello.com/c/gs2I3Cj2/676-end-your-search

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -854,3 +854,10 @@ class DataAPIClient(BaseAPIClient):
 
     def find_direct_award_project_services(self, user_id, project_id):
         raise NotImplementedError()
+
+    def lock_direct_award_project(self, user_email, project_id):
+        return self._post_with_updated_by(
+            "/direct-award/projects/{}/lock".format(project_id),
+            data={},
+            user=user_email,
+        )
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -3,12 +3,13 @@
 import re
 import six
 try:
-    from urllib.parse import urlparse, parse_qsl
+    from urllib.parse import urlparse, urlencode, urlunparse, parse_qs, parse_qsl
 except ImportError:
-    from urlparse import urlparse, parse_qsl
+    from urlparse import urlparse, urlunparse, parse_qs, parse_qsl
+    from urllib import urlencode
 
 
-from .base import BaseAPIClient
+from .base import BaseAPIClient, make_iter_method
 from .errors import HTTPError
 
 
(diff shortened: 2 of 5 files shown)


tests/test_apiclient.py
```diff
@@ -400,6 +400,15 @@ def test_get_aggregations_url(self, search_client):
     def test_get_index_from_search_api_url(self, search_client, search_api_url, expected_index):
         assert search_client.get_index_from_search_api_url(search_api_url) == expected_index
 
+    def test_search_service_from_url(self, search_client, rmock):
+        rmock.get(
+            'http://baseurl/g-cloud/services/search?lot=cloud-hosting',
+            json={'services': "myresponse"},
+            status_code=200)
+        result = search_client.search_services_from_url(search_api_url='https://baseurl/g-cloud/services/search'
```
allait: This might be a personal thing, but I find it's usually better to avoid the implicit string concatenation. Most of the time it looks like there's a missing comma between two arguments. At least that's how I'd read this code after a quick glance.

Something like this feels less ambiguous:

```
result = search_client.search_services_from_url(
    search_api_url='https://baseurl/g-cloud/services/search?lot=cloud-hosting'
)
```

dmapiclient/data.py
```diff
@@ -854,3 +854,14 @@ def get_direct_award_project_search(self, user_id, project_id, search_id):
 
     def find_direct_award_project_services(self, user_id, project_id):
         raise NotImplementedError()
+
+    def lock_direct_award_project(self, user_id, user_email, project_id):
+        return self._post_with_updated_by(
+            "/direct-award/projects/{}/lock".format(project_id),
+            data={
+                "project": {
+                    "userId": user_id
```
allait: This is probably a question for the API PR, but what's the user_id for?

tests/test_apiclient.py
samuelhwilliams: Reasonable. I'll change this.

dmapiclient/search.py
```diff
@@ -119,6 +119,19 @@ def search_services(self, index, q=None, page=None, **filters):
...
         return response
 
+    def search_services_from_url(self, search_api_url, page=None):
+        scheme, netloc, path, params, query, fragment = urlparse(search_api_url)
+        query_params = parse_qsl(query)
+        if page:
+            query_params.append(('page', page))
+        query = urlencode(query_params)
+
+        paged_search_api_url = urlunparse((scheme, netloc, path, params, query, fragment))
```
MuriloDalRi: Is it worth forking urllib.parse and implementing this if we're going to start doing this a lot?

All this parse, append, unparse, etc. is really tedious.

samuelhwilliams: Sensible, but not now. :trollface:

PR #93, opened 2017-09-07
Add `unlock_user` argument flag to `.update_user_password`

allait:
Allows unlocking the user and resetting the failed login count when
the password is changed.

Since password update is the same endpoint as the user update it makes
sense to do both changes in one API request to avoid any inconsistencies
that could be caused by one of the requests failing.

Updater details are the last argument for most API client methods, so
I'm adding a new argument before `updater`, making this a breaking
change. This shouldn't be an issue in practice since the method is
currently only used by the user-frontend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 Records breaking changes from major version bumps
 
+## 11.0.0
+
+PR: [#93](https://github.com/alphagov/digitalmarketplace-apiclient/pull/93)
+
+### What changed
+
+`DataApiClient.update_user_password` now has an additional `unlock_user` argument. When set to `True` it will unlock the user (clear the "5 failed login attempts") as well as reset the password.
+
+### Example app change
+
+Old
+```python
+api_client.update_user_password(1, 'newpassword', 'updater-email@example.com')
+```
+
+New
+
+```python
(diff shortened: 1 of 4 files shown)


TheDoubleK:
I don't know that we need a new flag for the endpoint?  Can we just make the API clear the failed login count to zero every time a password is reset?  I can't think why you'd ever want to reset the password but leave the account locked?

allait:
@TheDoubleK yeah, I can't think of a good use case for resetting the password without unlocking the user, but the API accepts both things at the moment and changing both if only "password" was set feels a bit weird for the generic update endpoint.

I was thinking of just adding `"locked": False` without an argument, but decided against it since it felt like it created implicit behaviour that's not mentioned in `update_user_password`. I made it default to `False` for the same reason - seeing the argument when the method is called should hopefully make it clear what's going on.

I could rename the method to `update_user_password_and_unlock` but I'm not sure if that's any better.

What do you think?

TheDoubleK:
It's probably fine either way.  Just adding unlocking to the existing API behaviour feels quicker and easier, and I think it's intuitive that if the system lets you reset a password then you should immediately be able to log in with that password.  So though it would be implicit behaviour I don't think it would be surprising.

But I don't think it makes any real difference either way.  :+1: if you like it this way round.

allait:
Closing this in favour of alphagov/digitalmarketplace-api#644

PR #95, opened 2017-09-11
Add "IDs only" option to search services

samuelhwilliams:
## Summary
When we lock Direct Award projects (at the 'end search' stage), we need to record all the service IDs that matched for the given search.
Bump version to 10.5.0.
Remove some 'helper' functions that exist for no real reason (get_search_url/get_aggregations_url)

## Ticket
https://trello.com/c/gs2I3Cj2/676-end-your-search

### DIFF
--- a/tests/test_apiclient.py
+++ b/tests/test_apiclient.py
@@ -363,6 +363,14 @@ class TestSearchApiClient(object):
         assert result == {'services': "myresponse"}
         assert rmock.last_request.query == 'page=10'
 
+    def test_search_services_id_only(self, search_client, rmock):
+        rmock.get(
+            'http://baseurl/g-cloud/services/search?idOnly=True',
+            json={'services': "myresponse"},
+            status_code=200)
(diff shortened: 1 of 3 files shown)


tests/test_apiclient.py
```diff
@@ -363,6 +363,15 @@ def test_search_services_with_pagination(self, search_client, rmock):
...
         assert rmock.last_request.query == 'page=10'
 
+    def test_search_services_id_only(self, search_client, rmock):
+        rmock.get(
+            'http://baseurl/g-cloud/services/search?idOnly=True',
+            json={'services': "myresponse"},
+            status_code=200)
+        result = search_client.search_services(index='g-cloud', id_only=True)
+        assert result == {'services': "myresponse"}
+        assert rmock.last_request.query == 'idonly=true'
```
galund: I suppose this must be case-insensitive, but seems weird to me that the "o" is not capitalized.

samuelhwilliams: It's case sensitive, but weirdly makes it lowercase...

PR #100, opened 2017-10-04
add tools for admin of service edits

risicle:
This includes the `user` filtering of `find_audit_events` and the addition of the `revert_service` method.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -570,6 +570,13 @@ class DataAPIClient(BaseAPIClient):
             user=user,
         )
 
+    def revert_service(self, service_id, archived_service_id, user):
+        return self._post_with_updated_by(
+            "/services/{}/revert".format(service_id),
+            data={"archivedServiceId": int(archived_service_id)},
+            user=user,
+        )
+
     def find_frameworks(self):
         return self._get("/frameworks")
 
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -25,29 +25,22 @@ class DataAPIClient(BaseAPIClient):
             user=None,
     ):
 
-        params = {}
+        params = {
+            "acknowledged": acknowledged,
+            "audit-date": audit_date,
+            "earliest_for_each_object": earliest_for_each_object,
+            "latest_first": latest_first,
+            "object-id": object_id,
+            "object-type": object_type,
+            "page": page,
+            "per_page": per_page,
+            "user": user,
+        }
+
         if audit_type:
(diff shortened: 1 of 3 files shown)


risicle:
```
./dmapiclient/data.py:14:5: C901 'DataAPIClient.find_audit_events' is too complex (12)
```
ha ha ha sod off.

risicle:
rrright now coveralls is complaining, but I suspect that's just because I've reduced the number of lines which were covered so the proportion went down.

dmapiclient/data.py
```diff
@@ -20,30 +20,26 @@ def find_audit_events(
             acknowledged=None,
             object_type=None,
             object_id=None,
+            user=None,
```
galund: This change would be more backward-compatible if the new argument was added last in the list?

galund:
Is there a ticket for this change?

dmapiclient/data.py
risicle: That it would, sir.

(though I would be more than a little worried if people were using >= 8 keyword arguments positionally)

samuelhwilliams: Technically API breaking change and should be a major version bump. :trollface:

tests/test_apiclient.py
```diff
@@ -591,6 +591,22 @@ def test_update_service_status(self, data_client, rmock):
...
+
+        data_client.revert_service(123, 314159, "jollypoldy@example.com")
+
+        assert tuple(req.json() for req in rmock.request_history) == (
+            {
+                "archivedServiceId": 314159,
+                "updated_by": "jollypoldy@example.com",
+            },
+        )
+
```
samuelhwilliams: The other tests seem to be prefer just checking `request_history[0]` directly, or `last_request`. I think this implementation confuses things by making you think about iterables when we know we're only interested in one call.

risicle: The other way doing it is fine as long as you also check `len(request_history) == 1`.

dmapiclient/data.py
```diff
@@ -21,29 +21,26 @@ def find_audit_events(
             object_type=None,
             object_id=None,
             latest_first=None,
-            earliest_for_each_object=None):
+            earliest_for_each_object=None,
+            user=None,
+    ):
+
+        params = {
+            "page": page,
```
TheDoubleK: The order of params here is pretty random, but also doesn't matter ¯\\_(ツ)_/¯

risicle: True. Alphabetized now.

PR #102, opened 2017-10-13
Creating Buyer email domains

katstevens:
Trello ticket: https://trello.com/c/JxwDhnDh/82-migrate-existing-domain-check-to-use-database

We need two new things from the API client: 
- a DataAPIClient method to call the new `/buyer-email-domain` API endpoint
- an AuditEvent type for creating a new BuyerEmailDomain object.

Refactoring:
- I split up the massive `test_apiclient` module as I was struggling to find the right place to put a new test. There is now one module each for `BaseAPIClient`, `SearchAPIClient` and `DataAPIClient`. There's a tiny bit of duplication around the `rmock` module-scope fixture, which is used in all three test modules (I will see if I can DRY this without too much hassle).
- The `DataAPIClient` module was still huge so I split it up into classes, based on whichever model the method is updating.

This is PR 2 out of 4. 
PR 1: Schema migration https://github.com/alphagov/digitalmarketplace-api/pull/672
PR 3: API model and route https://github.com/alphagov/digitalmarketplace-api/pull/674 
PR 4: Data migration https://github.com/alphagov/digitalmarketplace-api/pull/675


tests/test_search_api_client.py
```diff
@@ -0,0 +1,380 @@
...
+        ],
+        "serviceName": "My Iaas Service",
+        "serviceSummary": "IaaS Service Summary",
+        "serviceBenefits": [
+            "Free lollipop to every 10th customer",
+            "It's just lovely"
+        ],
+        "serviceFeatures": [
+            "[To be completed]",
+            "This is my second \"feture\""
```
MuriloDalRi: feture? 😛

tests/test_search_api_client.py
```diff
@@ -0,0 +1,380 @@
...
+        ],
+        "minimumContractPeriod": "Month",
+        "terminationCost": True,
+        "priceInterval": "",
+        "trialOption": True,
+        "priceUnit": "Person",
+        "educationPricing": True,
+        "vatIncluded": False,
+        "priceString": "£10.0067 per person",
+        "priceMin": 10.0067,
```
MuriloDalRi: Is that a valid price?

katstevens: No idea - I didn't change any of this, I just moved it 😄

MuriloDalRi: Yeah sorry, just realised it!

TheDoubleK: Yes that's a valid price.

For G-Cloud things you might have like "0.0067p per GB" or something.  There was a user need - in that suppliers complained *a lot* when we restricted it to 2 decimal places, I think during G-7 applications, and we had to release a fix for it quickly, during the application period.

TheDoubleK: e.g. this looks wrong, but is a valid pricing structure for G-Cloud: https://www.digitalmarketplace.service.gov.uk/g-cloud/services/911824080513785

MuriloDalRi: Ah yeah, that makes sense

dmapiclient/data.py
```diff
@@ -422,6 +422,15 @@ def is_email_address_with_valid_buyer_domain(self, email_address):
             "/users/check-buyer-email", params={'email_address': email_address}
         )['valid']
 
+    def create_buyer_email_domain(self, buyer_email_domain, user):
+        return self._post_with_updated_by(
+            "/buyer-email-domain",
+            data={
+                "buyerEmailDomains": {"domainName": buyer_email_domain}
```
TheDoubleK: Might we ever want to do a bulk update?  Probably not I guess, they only come in one at a time...

katstevens: Out of scope 😸

dmapiclient/audit.py
```diff
@@ -69,6 +69,9 @@ class AuditTypes(Enum):
     # Mailing list actions
     mailing_list_subscription = "mailing_list_subscription"
 
+    # Buyer Email Domains
+    create_buyer_email_domain = "create_buyer_email_domain"
```
TheDoubleK: Maybe this should sit under "Admin actions"?

katstevens: That is an excellent idea.

katstevens:
The test refactor was indeed all copy-paste 😸  Going to merge this now so Travis can pass on the API code PR.

PR #104, opened 2017-10-16
Deduplicate test fixture rmock into conftest file

benvand:
@katstevens Meant to comment about this on the PR but was too late. To make a fixture available to all tests just stick it in the `conftest.py`
https://docs.pytest.org/en/latest/fixture.html#sharing-a-fixture-across-tests-in-a-module-or-class-session

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,7 +1,24 @@
 import pytest
+import requests_mock
 from flask import Flask
 
 
 @pytest.fixture
 def app():
     return Flask(__name__)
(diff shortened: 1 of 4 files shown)


katstevens:
(Do we bother updating the version if we're just refactoring tests?)

benvand:
Good question, I forgot to be honest but actually yeah there's no real value in updating the version for this. The tests are only ever run by devs or travis. Devs will run locally and travis runs on commit.

PR #105, opened 2017-10-16
Update for saved search data tracking

Wynndow:
For this ticket: https://trello.com/c/eo8gJE4C

### Add `with-users` param to find_direct_award_projects  …
The serialize method for direct award projects has a `with-users`
parameter.

There is a PR up on the api to allow the `list_projects`
endpoint to accept `with-users` as a param

### Make `user-id` optional for direct-award endpoints  …
The endpoints on the API for direct award searches and services take an
optional `user-id` param.

The apiclient forced you to supply one. This made getting all searches
or services for a given project difficult.

### Fix `find_direct_award_project_searches_iter`  …
The `make_iter_method` method expects its second argument to be the name
of the model as returned in a serialized blob. For
direct_award_project_searches this is 'searches' and not 'projects'.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -851,7 +851,7 @@ class DataAPIClient(BaseAPIClient):
             params=params
         )
 
-    find_direct_award_project_searches_iter = make_iter_method('find_direct_award_project_searches', 'projects')
+    find_direct_award_project_searches_iter = make_iter_method('find_direct_award_project_searches', 'searches')
 
     def create_direct_award_project_search(self, user_id, user_email, project_id, search_url):
         return self._post_with_updated_by(
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -837,7 +837,7 @@ class DataAPIClient(BaseAPIClient):
             user=user_email
         )
 
-    def find_direct_award_project_searches(self, user_id, project_id, page=None, only_active=None):
+    def find_direct_award_project_searches(self, project_id, user_id=None, page=None, only_active=None):
         params = {
             "user-id": user_id,
             "page": page,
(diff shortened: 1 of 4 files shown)


dmapiclient/data.py
```diff
@@ -830,7 +834,7 @@ def create_direct_award_project(self, user_id, user_email, project_name):
             user=user_email
         )
 
-    def find_direct_award_project_searches(self, user_id, project_id, page=None, only_active=None):
+    def find_direct_award_project_searches(self, project_id, user_id=None, page=None, only_active=None):
```
risicle: ⚠️  API breaking change  ⚠️

dmapiclient/data.py
```diff
@@ -866,17 +870,21 @@ def get_direct_award_project_search(self, user_id, project_id, search_id):
             }
         )
 
-    def get_direct_award_project_services(self, user_id, project_id, fields=[]):
+    def get_direct_award_project_services(self, project_id, user_id=None, fields=[]):
```
risicle: ⚠️  API breaking change  ⚠️

dmapiclient/data.py
```diff
@@ -866,17 +870,21 @@ def get_direct_award_project_search(self, user_id, project_id, search_id):
...
+        params = {"user-id": user_id}
+        if fields:
+            params.update({"fields": ','.join(fields)})
```
risicle: Seems a little verbose over `params["fields"] = ",".join(fields)` but unimportant

Wynndow: `params["fields"] = ",".join(fields)` would set the `fields` param to `''` if it was `[]`

risicle: I wasn't suggesting you remove the `if fields:`

PR #110, opened 2017-11-07
Add new update-by-admin audit event type

TheDoubleK:
Part of this ticket: https://trello.com/c/DbYlqSxv/121-exclude-changes-made-by-admin-in-the-approve-edits-section

We need a new audit type to be able to easily distinguish edits by admins from edits by suppliers.

We need the API to take an additional (optional) parameter so it knows which type of audit event to use.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -547,9 +547,9 @@ class DataAPIClient(BaseAPIClient):
 
     find_services_iter = make_iter_method('find_services', 'services')
 
-    def update_service(self, service_id, service, user):
+    def update_service(self, service_id, service, user, by_admin=False):
         return self._post_with_updated_by(
-            "/services/{}".format(service_id),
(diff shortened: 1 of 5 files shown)


dmapiclient/data.py
```diff
@@ -547,9 +547,9 @@ def find_services(self, supplier_id=None, framework=None, status=None, page=None
 
     find_services_iter = make_iter_method('find_services', 'services')
 
-    def update_service(self, service_id, service, user):
+    def update_service(self, service_id, service, user, by_admin=False):
         return self._post_with_updated_by(
-            "/services/{}".format(service_id),
+            "/services/{}{}".format(service_id, "?by_admin=True" if by_admin else ""),
```
katstevens: Would something like `"?user_role=admin" if by_admin else ""` be more flexible?

TheDoubleK:
Updated as per conversation IRL.

PR #111, opened 2017-11-08
New `mapping` parameter for index creation.

galund:
Minor - not backward-compatible - change to the search API client, to account for the new `mapping` argument that is shortly arriving in the Search API. See <https://github.com/alphagov/digitalmarketplace-search-api/pull/115>.

https://trello.com/c/5oIKvS1s/100-search-service-for-briefs-plus-indexing-script

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 Records breaking changes from major version bumps
 
+## 12.0.0
+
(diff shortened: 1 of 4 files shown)


benvand:
Can we get and `OLD`/`NEW` in the changelog like the other entries? Makes finding the signature that's changed easier.

PR #112, opened 2017-11-15
Pass `object_type` to index function

Wynndow:
For this card on trello: https://trello.com/c/JR2Ch0ml

### Match object in reverse url search
The object type was hardcoded to 'services'. If it encounters a url with
a different object it won't match.

This allows it to match any object and return it

### Give `object_type` to `index`
It was decided that it would be more obvious what was happening if the
search-api client index method took the object type for the document
being indexed, and created the url using it.

The url is being constructed in the method instead of using the `_url`
method, as it's less work than giving `_url` a default, and is more
obvious.

### Don't match '/' when reversing url for index
Now that 'services' has been removed from the regex used to extract the
index from a search url, it would have matched any url regardless of how
deep it was.

By not matching a '/' we can limit it to a depth of three.


dmapiclient/search.py
```diff
@@ -98,9 +101,9 @@ def set_alias(self, alias_name, target_index):
             data={'type': 'alias', 'target': target_index}
         )
 
-    def index(self, index, service_id, service):
-        url = self._url(index, service_id)
-        return self._put(url, data={'service': service})
+    def index(self, index_name, object_type, object_id, serialized_object):
+        url = '/{}/{}/{}'.format(index_name, object_type, object_id)
+        return self._put(url, data={'document': serialized_object})
```
katstevens: Is there any value in defaulting the object_type to `service` for backwards compatibility (i.e. would this avoid the breaking change)?

Wynndow: Yeah, i did think about it, but then the change from 'service' to 'document' in the data is a breaking change anyway.

galund: Hmm, annoyingly in this case 'object type' has to be `services` with an ess - is possibly pedantry to mention it, but maybe it means we can think of another variable name. It's called type elsewhere, though.

Still wondering if having a default here might be a good idea, to keep backward compatibility?

The service->document change isn't breaking I think, in the sense that it's entirely captured here (i.e. clients of the library don't need to know about it).

risicle:
Interesting that you've used `object_type` - I've used `entity_name` so far, we should coordinate.

risicle:
(fwiw "object type" is kinda conflatible with the result of `type()`)

dmapiclient/search.py
Wynndow: @galund I'm a little confused about the '`services` with an ess' thing?

And I guess now that the search-api can handle 'document' or 'services' as the key it wouldn't actually break anything. 

I'll add the default and remove the breaking change.

galund: Get said "add 'service' for backward compat" and I was just pointing out that you have to add "service*s*" (unless I'm wrong!)

Wynndow: I think it'd need to be `service`, as that's what the search-api is currently looking for, and also what you're using in your current PR on the search-api?

Wynndow:
@risicle Could use `doc_type` as that's what the search-api is using?

dmapiclient/search.py
galund: Object type is being used in the URL - the "document" string is hard-coded (which is fine).

Wynndow: Yep, sorry, I'm with you, you're right. It was `services` that was being hardcoded into the url before.

risicle:
I _would_ be ok with `doc_type` except I've got to do my handling of it in the frameworks repo, where we haven't introduced the concept of "document" (and probably shouldn't have to).

galund:
I dunno, if you have a dep on the Search API, then you now have to acknowledge that it is now considering its incoming data to be documents, regardless of what you think they are, if u see what I mean :)

risicle:
don't really have a dep on the search api in the fw repo yet, but will go with `doc_type` anyway

galund:
It's a sorta-dep, in that you're producing a file in a format it knows how to consume. Whatevs :)

tests/test_search_api_client.py
```diff
@@ -77,6 +77,63 @@ def service():
     }
 
 
+@pytest.fixture
+def brief():
```
galund: Do these tests actually run against a Search API instance? If not, I don't really get how this fixture actually tests anything? (I get we're copying the G-Cloud fixture here; I don't understand that either.)

Wynndow: I think it's just to make the test as realistic as possible. In the tests, we're calling the method on the search_api_client, and passing in the variables it normally takes. In the tests that use this fixture, nothing is actually being done with it (it just goes into the data json), but it feels like we should test it with something as close to real as possible. In the future if someone changed the index method to do something different, it'd be good to have "real" brief in the test to catch any issues that might arise. Or something.

galund: Hmm, yeah I tend to think this is just yet another fixture that goes out of date as we iterate frameworks, in which case it might as well contain as little as possible, even if that makes it unrealistic. ISTM that the API _client_ should _never_ do anything much with the data it passes through (any important logic should be in the service)...

If we change this we should probably rip out the data from services fixture too. @risicle what you think?

risicle: Tend to agree unless we can think of any gotchas where we *are* dependent on some brief properties...?

Wynndow: Yep, that makes sense. I'll strip it out and look at taking the service out too.

PR #113, opened 2017-11-21
155 allow admin manager invite new admin users

benvand:
https://trello.com/c/2Uc2WMl2/155-allow-super-admin-invite-new-admin-users
https://docs.google.com/document/d/1MjvYx9aQWgcbryiaGrRLOPt_lVm8gIbqOH3vmkKx_fo/edit#

For this new page we agreed to restrict email addresses that can be invited to become an admin by domains listed in the app config.

* New apiclient method using new api end point
* Tests

### DIFF
--- a/tests/test_data_api_client.py
+++ b/tests/test_data_api_client.py
@@ -532,6 +532,29 @@ class TestBuyerDomainMethods(object):
         }
 
 
+class TestEmailVaildForAdminMethod(object):
+
+    def test_email_address_with_valid_admin_domain_is_true(self, data_client, rmock):
+        rmock.get(
+            "http://baseurl/users/valid-admin-email?email_address=kev%40gov.uk",
+            json={"valid": True},
+            status_code=200
+        )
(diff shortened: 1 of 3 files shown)


tests/test_data_api_client.py
```diff
@@ -532,6 +532,29 @@ def test_create_buyer_email_domain(self, data_client, rmock):
...
+        rmock.get(
+            "http://baseurl/users/valid-admin-email?email_address=kev%40gov.uk",
+            json={"valid": True},
+            status_code=200
+        )
+        result = data_client.email_is_valid_for_admin_user('kev@gov.uk')
+        assert rmock.called
+        assert result is True
+
+    def test_not_valid_api_response_is_false(self, data_client, rmock):
```
katstevens: Could we have more specific test names? The class above has `test_is_email_address_with_valid_buyer_domain_true`, so how about `test_is_email_address_with_valid_admin_domain_true` ?

benvand: Done 👍

PR #115, opened 2017-11-28
Add name parameter to 'update_user' method

CrystalPea:
This was done to enable editing admin user name as per this ticket:
https://trello.com/c/j1jx6KkX/154-allow-super-admin-edit-admin-roles

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -373,6 +373,7 @@ class DataAPIClient(BaseAPIClient):
                     active=None,
                     role=None,
                     supplier_id=None,
+                    name=None,
                     updater="no logged-in user"):
         fields = {}
         if locked is not None:
@@ -393,6 +394,11 @@ class DataAPIClient(BaseAPIClient):
         if supplier_id is not None:
             fields.update({
                 'supplierId': supplier_id
+            })
+
+        if name is not None:
+            fields.update({
+                'name': name
             })
 
         params = {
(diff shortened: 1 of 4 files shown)


dmapiclient/data.py
```diff
@@ -373,6 +373,7 @@ def update_user(self,
                     active=None,
                     role=None,
                     supplier_id=None,
+                    name=None,
```
TheDoubleK: If we add the parameter here it will count as a breaking change to version 13.0.0 and need an entry in the Changelog, as any code mad enough to call this with positional parameters will need to be updated.  It is definitely neater this way, and probably existing code would be fine as it should be using named params. 

If we don't want to break things though then `name` will have to come after `updater`, right at the end of the list. That way all existing code is guaranteed to continue to work.

Either way is OK with me, up to your which you prefer.

CrystalPea: Ok, I changed the version to 13.0.0, interesting how some small thing can be a breaking change sometimes 😮 . But makes total sense :D.

TheDoubleK:
Great. Please can you add a note in the `CHANGELOG.md` file to show people importing the new version what code changes they might need to make.

TheDoubleK:
And the test for this method probably also needs updating to check that it does the right thing with the new parameter.

CrystalPea:
Ahh, fair do @TheDoubleK. I did that now :). It's the first time I've introduced a breaking change 😳

PR #116, opened 2017-12-04
Search and aggregate briefs

Wynndow:
Part of this ticket: https://trello.com/c/KcqU7k20

### Set `doc_type` parameter for search functions
We want to be able to search for more than just services. Allowing
doc_type to be passed some functions will allow this. Setting the
default to `services` will maintain some backwards compatibility

### Add `search_briefs` method on search-api-client
This is really similar to `search_services`. The only reason for not
combining them is for backwards compatability. It may be something we
want to do.

### Generalize `aggregate_services` to `aggregate_docs`
We want to aggregate briefs as well as services now. I would have
created a new method `aggregate_briefs` to prevent the breaking change,
but this method is used by helper methods in the frontend and so needs
to be general.

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -20,17 +20,21 @@ from .exceptions import ImproperlyConfigured
 logger = logging.getLogger(__name__)
 
 
-def make_iter_method(method_name, model_name):
+def make_iter_method(method_name, *model_names):
     """Make a page-concatenating iterator method from a find method
 
     :param method_name: The name of the find method to decorate
-    :param model_name: The name of the model as it appears in the JSON response
+    :param model_names: The names of the possible models as they appear in the JSON response. The first found is used.
     """
     backoff_decorator = backoff.on_exception(backoff.expo, HTTPTemporaryError, max_tries=5)
 
     def iter_method(self, *args, **kwargs):
         method = getattr(self, method_name)
         result = backoff_decorator(method)(*args, **kwargs)
+        # Filter the list of model names for those that are a key in the response, then take the first.
+        # Useful for backwards compatability if response keys might change
+        model_name = next(filter(lambda name: name in result, model_names))
+
         for model in result[model_name]:
             yield model
 
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -33,7 +33,7 @@ def make_iter_method(method_name, *model_names):
         result = backoff_decorator(method)(*args, **kwargs)
         # Filter the list of model names for those that are a key in the response, then take the first.
         # Useful for backwards compatability if response keys might change
-        model_name = next(filter(lambda name: name in result, model_names))
+        model_name = next(model_name for model_name in model_names if model_name in result)
 
         for model in result[model_name]:
             yield model
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -157,7 +157,7 @@ class SearchAPIClient(BaseAPIClient):
             for service in result.get('documents') or result.get('services'):
                 yield service
 
-    def aggregate_docs(self, index, doc_type, q=None, aggregations=[], **filters):
+    def aggregate(self, index, doc_type, q=None, aggregations=[], **filters):
         response = self._get(
             self.get_url(path='aggregations', index=index, doc_type=doc_type, q=q, aggregations=aggregations, **filters)
         )
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -140,6 +140,6 @@ class SearchAPIClient(BaseAPIClient):
 
     search_services_from_url_iter = make_iter_method('search_services_from_url', 'services')
 
-    def aggregate_services(self, index, q=None, aggregations=[], **filters):
-        response = self._get(self.get_url(path='aggregations', index=index, q=q, aggregations=aggregations, **filters))
+    def aggregate_docs(self, index, doc_type, q=None, aggregations=[], **filters):
+        response = self._get(self.get_url(path='aggregations', index=index, doc_type=doc_type, q=q, aggregations=aggregations, **filters))
         return response
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -86,7 +86,7 @@ class SearchAPIClient(BaseAPIClient):
     def get_index_from_search_api_url(self, search_api_url):
         return self._url_reverse(search_api_url)[0]
 
-    def get_search_url(self, index, doc_type='services', q=None, page=None, **filters):
+    def get_search_url(self, index, q=None, page=None, doc_type='services', **filters):
         return self.get_url(path='search', index=index, doc_type=doc_type, q=q, page=page, **filters)
 
     def create_index(self, index, mapping):
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -2,6 +2,7 @@
 
 import re
 import six
+import backoff
 try:
     from urllib.parse import urlparse, urlencode, urlunparse, parse_qsl
 except ImportError:
@@ -9,8 +10,8 @@ except ImportError:
     from urllib import urlencode
 
 
-from .base import BaseAPIClient, make_iter_method
-from .errors import HTTPError
+from .base import BaseAPIClient
+from .errors import HTTPError, HTTPTemporaryError
 
 
 class SearchAPIClient(BaseAPIClient):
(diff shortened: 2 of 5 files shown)


dmapiclient/search.py
```diff
@@ -86,8 +86,8 @@ def get_frontend_params_from_search_api_url(self, search_api_url):
     def get_index_from_search_api_url(self, search_api_url):
         return self._url_reverse(search_api_url)[0]
 
-    def get_search_url(self, index, q=None, page=None, **filters):
-        return self.get_url(path='search', index=index, q=q, page=page, **filters)
+    def get_search_url(self, index, doc_type='services', q=None, page=None, **filters):
+        return self.get_url(path='search', index=index, doc_type=doc_type, q=q, page=page, **filters)
```
risicle: You should either append the new `doc_type` arg onto the "end" of the args or add it as a breaking change to `CHANGELOG` as it could in theory be being used positionally by clients. (yeah, I know...)

dmapiclient/search.py
```diff
@@ -136,6 +144,8 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
 
     search_services_from_url_iter = make_iter_method('search_services_from_url', 'services')
 
-    def aggregate_services(self, index, q=None, aggregations=[], **filters):
-        response = self._get(self.get_url(path='aggregations', index=index, q=q, aggregations=aggregations, **filters))
+    def aggregate_docs(self, index, doc_type, q=None, aggregations=[], **filters):
```
risicle: same note here about arg order.

tests/test_search_api_client.py
```diff
@@ -179,47 +179,87 @@ def test_search_services(self, search_client, rmock):
...
+    def test_search_with_pagination(self, search_client, rmock, doc_type, index):
         rmock.get(
-            'http://baseurl/g-cloud/services/search?page=10',
-            json={'services': "myresponse"},
+            'http://baseurl/{}/{}/search?page=10'.format(index, doc_type),
+            json={'documents': "myresponse"},
             status_code=200)
-        result = search_client.search_services(index='g-cloud', page=10)
-        assert result == {'services': "myresponse"}
+        result = eval("search_client.search_{}(index='{}', page=10)".format(doc_type, index))
```
risicle: Yarg!

You can always do
```
search_client.getattr("search_{}".format(doc_type))(index, page=10)
```

dmapiclient/search.py
Wynndow: This is a breaking change because of the change in method name though? And the new argument is at the end of the positional args anyway?

risicle: I am clearly unable to read.

risicle: (but also, "keyword" args *can* be called positionally)

Wynndow:
@risicle I've pushed a new commit to allow the api to not break with the search-api changes.

dmapiclient/search.py
```diff
@@ -134,8 +143,28 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
...
+
+        while True:
+            if 'next' not in result['links']:
+                return
+
+            result = backoff_decorator(self._get)(result['links']['next'])
+            for service in result.get('documents') or result.get('services'):
+                yield service
+
+    def aggregate_docs(self, index, doc_type, q=None, aggregations=[], **filters):
```
galund: Can we just call it "aggregate"? aggregate_docs sounds a bit weird, I know it's consistent with how it used to be, but meh...

dmapiclient/search.py
```diff
@@ -142,7 +143,25 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
 
         return self._get(paged_search_api_url)
 
-    search_services_from_url_iter = make_iter_method('search_services_from_url', 'services')
+    # search_services_from_url_iter = make_iter_method('search_services_from_url', 'documents')
+
+    # This is a temporary method to allow for backwards compatability while switching from the search-api returning
+    # `documents` instead of just `services`. The data-api uses this method when locking direct award searches. The api
+    # will need to pull in this apiclient before the search-api changes. Once it does, this method can be removed and
+    # the calle to `make_iter_method` above can be uncommented.
```
risicle: > `calle`

typo

dmapiclient/search.py
```diff
@@ -116,7 +117,15 @@ def delete(self, index, service_id):
         return None
 
     def search_services(self, index, q=None, page=None, id_only=False, **filters):
-        response = self._get(self.get_search_url(index=index, q=q, page=page, id_only=id_only, **filters))
+        response = self._get(
+            self.get_search_url(index=index, doc_type='services', q=q, page=page, id_only=id_only, **filters)
+        )
+        return response
+
+    def search_briefs(self, index, q=None, page=None, id_only=False, **filters):
```
galund: The aggregate function (only one, you pass in the type as a parameter) is now inconsistent with the search functions (there are two, the type is based on which function you call).

I think I prefer just having the one function in the library (i.e. maybe we could have "search" rather than "search_services" and "search_briefs".

And do we not need an 'iter' version of the briefs search?

dmapiclient/search.py
```diff
@@ -134,8 +143,28 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
...
+        result = backoff_decorator(self.search_services_from_url)(search_url, id_only=id_only)
+        for service in result.get('documents') or result.get('services'):
+            yield service
+
+        while True:
+            if 'next' not in result['links']:
+                return
+
+            result = backoff_decorator(self._get)(result['links']['next'])
+            for service in result.get('documents') or result.get('services'):
```
galund: I guess this is a temporary thing, but it feels like less code to do something to `make_iter_method` to allow it to look for more than one thing (kwargs maybe?)

dmapiclient/search.py
```diff
@@ -142,7 +143,25 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
...
+    # search_services_from_url_iter = make_iter_method('search_services_from_url', 'documents')
+
+    # This is a temporary method to allow for backwards compatability while switching from the search-api returning
+    # `documents` instead of just `services`. The data-api uses this method when locking direct award searches. The api
+    # will need to pull in this apiclient before the search-api changes. Once it does, this method can be removed and
+    # the calle to `make_iter_method` above can be uncommented.
+    def search_services_from_url_iter(self, search_url, id_only):
+        backoff_decorator = backoff.on_exception(backoff.expo, HTTPTemporaryError, max_tries=5)
+        result = backoff_decorator(self.search_services_from_url)(search_url, id_only=id_only)
+        for service in result.get('documents') or result.get('services'):
```
risicle: In the case where `documents` were a zero-length sequence, that being Falsey would cause it to attempt to `.get(` `services` which would of course be absent, leading the `for` loop to try and iterate through the `None` it got from that, raising an exception.

Could solve by adding an `or ()` to the end of the chain or modifying the last call to `result.get('services', ())`.

dmapiclient/search.py
```diff
@@ -142,7 +143,25 @@ def search_services_from_url(self, search_api_url, id_only=False, page=None):
...
+            yield service
+
+        while True:
+            if 'next' not in result['links']:
+                return
+
+            result = backoff_decorator(self._get)(result['links']['next'])
+            for service in result.get('documents') or result.get('services'):
```
risicle: same here

dmapiclient/search.py
risicle: (but yes, of course, moving your arg to the end wouldn't allow it to be required)

dmapiclient/search.py
risicle: (thanks though, this should make my job easier)

dmapiclient/search.py
Wynndow: Yeah, this makes sense. I think I was keeping them separate for backwards compatibility, but it should be fine.

And no, no need for a briefs iter.

dmapiclient/search.py
Wynndow: Yep, all done.

dmapiclient/search.py
Wynndow: Yep, have done a thing with `make_iter_method`'s *args to allow it to take more than one model name. The iter method will used the first model name that it finds in the response.

dmapiclient/search.py
Wynndow: Have removed this now and update `make_iter_method` to do something different. Would be good to get your eyes on it.

dmapiclient/base.py
```diff
@@ -20,17 +20,21 @@
...
+    :param model_names: The names of the possible models as they appear in the JSON response. The first found is used.
     """
     backoff_decorator = backoff.on_exception(backoff.expo, HTTPTemporaryError, max_tries=5)
 
     def iter_method(self, *args, **kwargs):
         method = getattr(self, method_name)
         result = backoff_decorator(method)(*args, **kwargs)
+        # Filter the list of model names for those that are a key in the response, then take the first.
+        # Useful for backwards compatability if response keys might change
+        model_name = next(filter(lambda name: name in result, model_names))
```
galund: `next(name for name in model_names if name in result)` is slightly more vogue, but whatever :)

Wynndow: It's also py2 compliant, which mine isn't! Changing!

PR #120, opened 2018-02-07
Filter brief responses by awarded_at date

brenetic:
`awarded_at` can now be passed into the api as a date in the form of `YYY-MM-DD`.

Merge after:
- [x] https://github.com/alphagov/digitalmarketplace-api/pull/735

https://trello.com/c/LsEbL3jP

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '14.0.1'
+__version__ = '14.1.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Maybe squash the commits, too.

samuelhwilliams:
We don't need separate commits for version bumps any more. Does the manual still say we should be doing that? 🤔

katstevens:
Good to know - I don't think it's mentioned specifically anywhere in the manual? I think a separate commit is still needed for at least one of the non-Python libraries, let me see... https://github.com/alphagov/digitalmarketplace-frameworks?

PR #122, opened 2018-02-16
Filter Briefs by unusccessful_on and cancelled_on dates

katstevens:
We can already filter Briefs by these date fields on the API route. This PR just adds the kwargs to the API client method. 

Usage:
```
result = data_client.find_briefs(cancelled_on="2017-10-20")
result = data_client.find_briefs(unsuccessful_on="2017-10-20")
```


katstevens:
The `find_foo_iter` methods are useful but I agree it could do with a revamp (firebreak maybe! 🔥 🚒 )

PR #123, opened 2018-02-19
Rename method, fix tests

samuelhwilliams:
Remove get_direct_award_project_services_iter in favour of find_x equivalent; fix some tests to use _test_find_iter correctly

## Ticket
https://trello.com/c/QV36xkLW/216-getmodeliter-v-findmodeliter-in-the-apiclient

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 Records breaking changes from major version bumps
 
+## 15.0.0
+
+PR: [#123](https://github.com/alphagov/digitalmarketplace-apiclient/pull/123)
+
+Removes `get_direct_award_project_services_iter` in favour of `find_direct_award_project_services_iter`, which is the
+more consistent naming scheme.
+
+Old:
+```python
+data_api_client.get_direct_award_project_services_iter(...)
+```
+
+New:
+```python
+data_api_client.find_direct_award_project_services_iter(...)
+```
+
 ## 14.0.0
 
 PR: [#116](https://github.com/alphagov/digitalmarketplace-apiclient/pull/116)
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
@Wynndow no idea what you're talking about 🙈

Wynndow:
Seems a bit harsh of Travis...

samuelhwilliams:
😨

samuelhwilliams:
I'm going to suggest we ignore coveralls in this instance. I think coverage has dropped because we had 3 lines of code covered and have now reduced that to 2, which results in the overall percentage dropping (because the number of lines of untested code has remained constant, but the number of lines of tested code has dropped). Go Coveralls.

If you agree with my logic, please can you approve @Wynndow, and I'll override Coveralls. Thanks. 👍

Wynndow:
Yeah, I'm with you. Can't we configure Coveralls to only protest if coverage drops be a certain percent or below a hard limit? Might be worth looking at that?

samuelhwilliams:
We can. On all other repos we have it set to only block us if coverage drops by 0.5% of more. This is the only one where it's set to 0% threshold. It's probably ok to change it to match the other ones.

Wynndow:
Agreed.

risicle:
@samuelhwilliams oh dear. I think you forgot to actually bump the version in this PR and now we're forging ahead with `v14.8.`something without realizing it has a breaking change between `14.4.0` and `14.5.0`

I guess I should apply a `15.0.0` bump asap.

risicle:
@MuriloDalRi please be aware of this before trying to propagate your change thinking it's only a minor bump.

samuelhwilliams:
🤦‍♂️

Annoying. Sorry @risicle. Thanks for updating.

risicle:
Actually kinda fortunate, lead me to this issue https://trello.com/c/imymWr7B/387-insufficient-mocking-of-dataapiclient-in-buyer-frontend-tests-could-be-dangerous

PR #128, opened 2018-03-08
request id handling: switch to using request.get_onwards_request_headers if available

risicle:
...however still fall back to using `request.request_id` failing that for compatibility: I *think* I've upgraded the `-utils` on all the relevant apps but could have missed something, so playing safe.

Strikes me that there isn't really any sort of integration test for this functionality - checking that a header makes it all the way through. Not sure where such a thing would live.

Trello https://trello.com/c/exuNFb7A/357-zipkin-headers-pt2-api-client , part of the broader https://trello.com/c/UOXf4vZZ/308-make-use-of-zipkin-headers-in-preference-to-request-id

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -113,7 +113,12 @@ class BaseAPIClient(object):
             "Authorization": "Bearer {}".format(self.auth_token),
             "User-agent": "DM-API-Client/{}".format(__version__),
         }
-        headers = self._add_request_id_header(headers)
+        if has_request_context():
+            if callable(getattr(request, "get_onwards_request_headers", None)):
+                headers.update(request.get_onwards_request_headers())
+            elif getattr(request, "request_id", None) and current_app.config.get("DM_REQUEST_ID_HEADER"):
+                # support old .request_id attr for compatibility
+                headers[current_app.config["DM_REQUEST_ID_HEADER"]] = request.request_id
 
         start_time = monotonic()
         try:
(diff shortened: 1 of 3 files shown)


dmapiclient/base.py
```diff
@@ -113,7 +113,12 @@ def _request(self, method, url, data=None, params=None):
             "Authorization": "Bearer {}".format(self.auth_token),
             "User-agent": "DM-API-Client/{}".format(__version__),
         }
-        headers = self._add_request_id_header(headers)
+        if has_request_context():
+            if callable(getattr(request, "get_onwards_request_headers", None)):
+                headers.update(request.get_onwards_request_headers())
+            elif getattr(request, "request_id", None) and current_app.config.get("DM_REQUEST_ID_HEADER"):
+                # support old .request_id attr for compatibility
```
idavidmcdonald: I wonder if worth putting this as a TODO to remove? Or will you be doing that very imminently?

risicle: Hopefully the latter, but would be good to do a collective think over whether I've missed anywhere...

risicle:
Yeah well we have vaguely thrown around the idea of being able to make assertions about log output from the functional tests in the past, but we didn't end up needing it because you did your thing with the alert. I think we concluded that the whole credentials thing (letting the FTs have access to the logs somehow) would be too much of a nightmare.

risicle:
I don't really think the tests need be more complex or there's really more that can go wrong in the more extensible implementation. If anything, this approach really just makes "our" code a fairly straightforward use of `dict.update()`. This code is what I would call "thinner".

PR #131, opened 2018-04-11
*_iter methods: give these all correct __name__s

risicle:
This fell out of a 2ndline ticket https://trello.com/c/f8SAV9WL/261-failure-of-data-export-job-overnight - turns out it's kinda annoying to look at tracebacks without "correct" function names.

This isn't a very pretty way of doing it but is the least complex way of doing it that doesn't introduce significant possibilities of wrongness.

The slightly more automatic way of doing this would involve maybe using a class decorator to do the name assignment post-declaration.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -53,6 +53,7 @@ class DataAPIClient(BaseAPIClient):
         )
 
     find_audit_events_iter = make_iter_method('find_audit_events', 'auditEvents')
+    find_audit_events_iter.__name__ = str("find_audit_events_iter")
 
     def acknowledge_audit_event(self, audit_event_id, user):
         return self._post_with_updated_by(
@@ -107,6 +108,7 @@ class DataAPIClient(BaseAPIClient):
         )
 
     find_suppliers_iter = make_iter_method('find_suppliers', 'suppliers')
+    find_suppliers_iter.__name__ = str("find_suppliers_iter")
 
     def get_supplier(self, supplier_id):
         return self._get(
@@ -288,6 +290,7 @@ class DataAPIClient(BaseAPIClient):
         )
 
     find_framework_suppliers_iter = make_iter_method('find_framework_suppliers', 'supplierFrameworks')
+    find_framework_suppliers_iter.__name__ = str("find_framework_suppliers_iter")
 
     # Users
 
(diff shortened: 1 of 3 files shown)


samuelhwilliams:
Actually, there's a Python builtin for doing this that we should be using - `functools.wraps`.

https://docs.python.org/3/library/functools.html#functools.wraps

risicle:
`__wraps__` isn't really the right thing here, because we don't have an existing function we want the new function to take on the identity of...

samuelhwilliams:
Mmm fair point, maybe.

Though it still feels like a simpler place for this is in the definition of `make_iter_method`. You could monkey-patch the name of the method at that point I think - and just suffix '_iter' on the passed-in method name. Maybe?

dmapiclient/data.py
```diff
@@ -53,6 +53,7 @@ def get_audit_event(self, audit_event_id):
         )
 
     find_audit_events_iter = make_iter_method('find_audit_events', 'auditEvents')
+    find_audit_events.__name__ = str("find_audit_events_iter")
```
samuelhwilliams: Wrong object being edited, if you keep this way of doing things.

risicle: huh. think my regex went awry. well caught.

risicle:
So yeah I considered doing the "internally renamed" thing but didn't because it doesn't *know* what you're going to assign that final function to, and it would be hyper confusing if it got it wrong because the user just happened to be doing something slightly different in some specific case.

samuelhwilliams:
That's certainly true, but do we think it's _likely_ a developer is going to create an iterator using this `make_iter_method` and then name is something completely wrong/weird/inconsistent?

risicle:
My rule of thumb is - if I did something I would expect to be able to do and it blows up in my face, is this a situation where I would be cursing and saying "who the hell wrote this piece of naiveity?" and in this case I think it would.

samuelhwilliams:
Seems reasonable.

Why do we need the `str()` bits around the names?

risicle:
I put a comment in the commit - it's because this file uses `from __future__ import unicode_literals` and py2 gets unhappy if you try and assign a unicode string to `__name__` - both py2 and py3 want their respective "native" `str`s

samuelhwilliams:
Ah man, I forgot this was still on Python 2.

Not sure it needs to be, but I guess that's one for another day.

galund:
Have created ticket for this https://trello.com/c/DlKHUT4a/430-remove-python-2

PR #132, opened 2018-04-12
Add method for copying all lots services

Wynndow:
Part of this trello ticket: https://trello.com/c/Nbv7N4lf

### Add method for copying all lots services
This is for calling the endpoint on the api for copying all of a
particular lots services between frameworks.

### Switch from flake8-putty to flake8-per-file-ignores
To support newer Python3 syntax through flake8 3.5.0. No per file
ignores are required currently but installing it keeps consistency with
how we use it elsewhere.

### DIFF
--- a/tests/test_data_api_client.py
+++ b/tests/test_data_api_client.py
@@ -1116,13 +1116,33 @@ class TestDraftServiceMethods(object):
         )
 
         result = data_client.copy_draft_service_from_existing_service(
-            2, 'user'
+            2, 'user', {'some': 'data'}
         )
 
         assert result == {"done": "it"}
         assert rmock.called
         assert rmock.request_history[0].json() == {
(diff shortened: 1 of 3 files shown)


tests/test_data_api_client.py
```diff
@@ -1116,13 +1116,33 @@ def test_copy_draft_service_from_existing_service(
...
         assert rmock.request_history[0].json() == {
-            'updated_by': 'user'
+            'updated_by': 'user',
+            'some': 'data',
+        }
+
+    def test_copy_published_from_framework(
+            self, data_client, rmock):
+        rmock.post(
+            "http://baseurl/draft-services/dos-cloud/sausages/copy-published-from-framework",
```
samuelhwilliams: 😂

PR #141, opened 2018-06-07
Add new outcome calls

risicle:
Trello https://trello.com/c/N6Np8b1F/126-apiclient-additions-for-new-backend-endpoints

Striaghtforward enough right?


risicle:
Blast - this is failing because it's using f-strings and this repo still tests against py2. But f-strings are particularly useful for this repo. Are we at the stage where we can abandon py2 for this repo?

risicle:
Sod it I've changed it to use `.format()`.

PR #142, opened 2018-06-18
208 delete users code that takes the user and strips all the personal data out

benvand:
https://github.com/alphagov/digitalmarketplace-api/pull/802

This PR:
* Bumps the version
* Adds a new AuditType for removing the personal data from a user
* Adds methods to interact with the `remove-personal-data` user and contact information api endpoints
* Adds tests for these methods

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '16.3.0'
+__version__ = '16.4.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


katstevens:
Weird... where did the commit with the AuditType disappear to?

benvand:
Sam suggested using the existing audit types for update user and update contact and as these models shouldn't change after the personal data has been removed then we can discern which audit event removed the personal data (the last update_user/ update_contact audit event)

katstevens:
Cool - makes sense 👍

PR #143, opened 2018-06-20
BaseAPIClient: add childSpanId parameter to log output

risicle:
This required a bit of tidying & refactoring to be doable in a clear, extensible and robust way.

Also add tests for the presence/absence of this childSpanId parameter, incidentally the first tests of logging behaviour for this library.

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -102,12 +102,6 @@ class BaseAPIClient(object):
             return None
 
         url = self._build_url(url, params)
-
-        logger.debug("API request {method} {url}",
-                     extra={
-                         'method': method,
-                         'url': url
-                     })
         headers = {
             "Content-type": "application/json",
             "Authorization": "Bearer {}".format(self.auth_token),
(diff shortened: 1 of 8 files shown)


dmapiclient/base.py
```diff
@@ -120,11 +114,41 @@ def _request(self, method, url, data=None, params=None):
...
+        child_span_id = next(
+            (
+                ci_headers[header_name] for header_name in (current_app.config.get("DM_SPAN_ID_HEADERS") or ())
+                if ci_headers.get(header_name)
+            ),
+            None,
+        ) if has_request_context() else None
+
+        common_log_extra = {
+            **({"childSpanId": child_span_id} if child_span_id is not None else {}),
```
katstevens: This line is failing the Python 2 flake8 check - I think we can safely take the 2.7 out of the travis.yml for this repo now?

dmapiclient/base.py
```diff
@@ -120,11 +114,41 @@ def _request(self, method, url, data=None, params=None):
...
+        del headers
+
+        # determine our final outgoing span id - find the first of DM_SPAN_ID_HEADERS which is set to something truthy
+        child_span_id = next(
+            (
+                ci_headers[header_name] for header_name in (current_app.config.get("DM_SPAN_ID_HEADERS") or ())
+                if ci_headers.get(header_name)
+            ),
+            None,
+        ) if has_request_context() else None
```
katstevens: Is it worth bothering testing the scenario where `has_request_context is None` ?  (I don't think the scenarios below cover it...)

dmapiclient/base.py
katstevens: Though officially removing Python 2 support would probably mean a major bump. Perhaps for another time.

risicle:
On Thursday, 21 June 2018 08:54:30 BST Kat wrote:

> 
> Is it worth bothering testing the scenario where `has_request_context is
> None` ?  (I don't think the scenarios below cover it...)

Probably yes - will squeeze one in.

dmapiclient/base.py
risicle: People suggested I just major-bump anyway. Done now.

PR #149, opened 2018-06-27
Add 'we are still assessing' method

lfdebrux:
This PR adds the apiclient method needed for the 'we are still assessing flow' [ticket].

It also adds a `_patch` method to the base api client class, so that the new PATCH endpoint from https://github.com/alphagov/digitalmarketplace-api/pull/807 can be used.

[ticket]: https://trello.com/c/5Lrp1lOK/110-fe-8-we-are-still-assessing-flow

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '17.2.0'
+__version__ = '17.3.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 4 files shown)


dmapiclient/__init__.py
```diff
@@ -1,4 +1,4 @@
-__version__ = '17.2.0'
+__version__ = '17.4.0'
```
katstevens: Did we miss `17.3.0`?

lfdebrux: I bumped it twice, once when adding `_patch` and once when adding `mark_direct_award_still_assessing`. I can change it to one bump though.

PR #150, opened 2018-07-04
Add generic update method

galund:
Add generic update method: the API's method is intended to be generic, so logically (arguably) the API Client should be fairly generic too.

Goes with https://github.com/alphagov/digitalmarketplace-api/pull/814

https://trello.com/c/7wlBbGU0/138-add-step-3-assessing-your-service-and-badges


galund:
(Test added, thanks, forgot :) )

PR #151, opened 2018-07-05
Log more detail on request exceptions

samuelhwilliams:
## Summary
We have been getting some mystery 503s in the buyer-frontend and don't
get any useful error information from the apiclient. This update causes
the apiclient to emit more useful information via the str() and repr()
calls on the base error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 18.0.0
+
+PR: [#151](https://github.com/alphagov/digitalmarketplace-apiclient/pull/151)
+
+The default request error message for unrecognised/unhandled errors has changed from 'Request failed' to 'Unknown request failure in dmapiclient'. 'Request failed' should be seen only in exceptional circumstances - most exceptions raised from failed requests should now hold the str() and repr() calls of the request error as their message.
+
+Any code that relies on the specific format of the request errors raised by the apiclient will need to be updated.
+
 ## 17.0.0
 
 PR: [#143](https://github.com/alphagov/digitalmarketplace-apiclient/pull/143)
--- a/tests/test_base_api_client.py
+++ b/tests/test_base_api_client.py
@@ -4,6 +4,7 @@ import logging
 
 from flask import request
 import requests
+import requests_mock
 import pytest
 import mock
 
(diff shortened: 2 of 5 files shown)


tests/test_base_api_client.py
```diff
@@ -53,7 +53,8 @@ def test_http_error_raises_api_error(self, base_client, rmock):
         with pytest.raises(HTTPError) as e:
             base_client._request("GET", '/')
 
-        assert e.value.message == REQUEST_ERROR_MESSAGE
+        assert e.value.message == "500 Server Error: None for url: http://baseurl/\n" \
+                                  "HTTPError('500 Server Error: None for url: http://baseurl/',)"
```
risicle: Perhaps this isn't under your control here, but `None for url:...` isn't the most enlightening message to get.

samuelhwilliams: Yeah, this I don't think is under my control. It is at least a more descriptive error than what we currently get though ('Request failed'), which is the root goal here.

CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 
 Records breaking changes from major version bumps
 
+## 18.0.0
+
+PR: [#151](https://github.com/alphagov/digitalmarketplace-apiclient/pull/151)
+
+The default request error message for unrecognised/unhandled errors has changed from 'Request failed' to 'Unknown request failure in dmapiclient'. 'Request failed' should be seen only in exceptional circumstances - most exceptions raised from failed requests should now hold the str() and repr() calls of the request error as their message.
+
+Any code that relies on the specific format of the request errors raised by the apiclient will need to be updated.
```
risicle: This is one of those "as someone who doesn't know the whole of `digitalmarketplace-foo-frontend` like the back of my hand, *how the hell* am I expected to check for this?" lines... not your fault though, just an issue with the whole "not my problem (except it actually is)" philosophy of our architecture.

risicle:
@katstevens though we should be careful this is exactly the sort of time he'd be trying to insert easter eggs for us.

CHANGELOG.md
samuelhwilliams: Yeah I mean I understand what you're saying. Really not sure how I'm meant to give more detailed instructions in this case, though...

PR #152, opened 2018-07-19
Add retry to base api client requests

Wynndow:
See [this related PR on utils to add logging for this.](https://github.com/alphagov/digitalmarketplace-utils/pull/427)

This was heavily influenced by this blog post: [https://www.peterbe.com/plog/best-practice-with-retries-with-requests](https://www.peterbe.com/plog/best-practice-with-retries-with-requests)

This will allow the api client to retry requests which raise connection
errors, read errors, as well as requests that return certain 500 error codes.

For the tests, I've had to reach quite a long way in to the requests and
urllib3 modules. This is because response_mock (rmock fixture) mocks the
adapter that requests uses. Doing this means that we never actually
retry anything. I was torn between doing this potentially quite fragile
testing method, or just testing that Retry was called correctly when a
request was made. I am very open to changing this if people think this is a
bad idea.

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -11,11 +11,10 @@ from requests.adapters import HTTPAdapter
 from requests.packages.urllib3.util.retry import Retry
 from flask import has_request_context, request, current_app
 
-import backoff
 from monotonic import monotonic
 
 from . import __version__
-from .errors import APIError, HTTPError, HTTPTemporaryError, InvalidResponse
+from .errors import APIError, HTTPError, InvalidResponse
 from .exceptions import ImproperlyConfigured
 
 
@@ -28,11 +27,8 @@ def make_iter_method(method_name, *model_names):
     :param method_name: The name of the find method to decorate
     :param model_names: The names of the possible models as they appear in the JSON response. The first found is used.
     """
-    backoff_decorator = backoff.on_exception(backoff.expo, HTTPTemporaryError, max_tries=5)
-
     def iter_method(self, *args, **kwargs):
-        method = getattr(self, method_name)
-        result = backoff_decorator(method)(*args, **kwargs)
+        result = getattr(self, method_name)(*args, **kwargs)
         # Filter the list of model names for those that are a key in the response, then take the first.
         # Useful for backwards compatability if response keys might change
         model_name = next(model_name for model_name in model_names if model_name in result)
@@ -44,7 +40,7 @@ def make_iter_method(method_name, *model_names):
             if 'next' not in result.get('links', {}):
                 return
 
-            result = backoff_decorator(self._get)(result['links']['next'])
+            result = self._get(result['links']['next'])
             for model in result[model_name]:
                 yield model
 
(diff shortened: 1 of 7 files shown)


dmapiclient/base.py
```diff
@@ -50,6 +52,10 @@ def iter_method(self, *args, **kwargs):
 
 
 class BaseAPIClient(object):
+    RETRIES = 3
+    RETRIES_BACKOFF_FACTOR = 0.3
+    RETRIES_FORCE_STATUS_CODES = (500, 502, 503, 504)
```
risicle: A comment to explain what `RETRIES_FORCE_STATUS_CODES` does?

Wynndow: Yep, I'll add one. It's a list of HTTP status codes on which to force a retry. Default is none.

tests/test_base_api_client.py
```diff
@@ -33,29 +36,70 @@ def _empty_context_manager():
...
-        rmock.request(
-            "GET",
-            "http://baseurl/",
-            text="Internal Error",
-            status_code=500)
+    @mock.patch('urllib3.connectionpool.HTTPConnectionPool._make_request')
+    @mock.patch('dmapiclient.base.BaseAPIClient.RETRIES_BACKOFF_FACTOR', 0)
+    def test_client_retries_on_read_error_and_raises_api_error(self, _make_request, base_client):
+        _make_request.side_effect = ProtocolError(mock.Mock(), 'Im a message')
+        retry_count = randint(1, 3)
```
risicle: Rather than introducing randomness & possibly random failures, you could just parametrize across all these values.

Wynndow: Yeah, i thought this might be nice way of not having to run the test multiple times each test run though. What possible random failures do you foresee?

risicle: Well just that if someone breaks the 3-retry case it will only fail one in three times -> could make it thru travis, then start giving people mysterious test failures. Cue much "what have I done??" head scratching.

risicle: (that said, I'm also guilty of using random in tests, but I'm considering removing it sometime)

dmapiclient/base.py
```diff
@@ -104,6 +111,22 @@ def _build_url(self, url, params):
...
+            connect=self.RETRIES,
+            status=self.RETRIES,
+            backoff_factor=self.RETRIES_BACKOFF_FACTOR,
+            status_forcelist=self.RETRIES_FORCE_STATUS_CODES,
+            raise_on_status=False,
+        )
+        adapter = HTTPAdapter(max_retries=retry)
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
+        return session
```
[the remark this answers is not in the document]
Wynndow: So I think that they will both apply. The backoff_decorator will try a max of 5 times, the Retry will try again 3 times. I think that means there could be up to 20 attempts if things don't work...

I could remove backoff from the iter method, or leave as is?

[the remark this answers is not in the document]
katstevens: Not sure. I'm starting to think this should be a breaking change - some of our scripts have their own retries (not sure about the FE apps), and we should adjust those where necessary.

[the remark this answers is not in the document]
Wynndow: I've just tested it and my guess above was right. It's pretty straightforward to remove backoff from the iter method and we still get the retries. I could up up the count to 5 to match the back_off config.

And yeah, I think you're right about a breaking change. I don't necessarily think it'll break anything, as all the retrying is done on the urllib3 side of thing and we ultimatly just get a requests Response object back, but if scripts stuff have their own back_off a breaking change makes sense.

PR #154, opened 2018-08-13
Add data_supplier_id param to `find_audit_events`

Wynndow:
For [this trello ticket.](https://trello.com/c/AdtcNmBR)
See [this associated PR on the api.
](https://github.com/alphagov/digitalmarketplace-api/pull/826)

The api now accepts this parameter to filter audit events by the
supplierId field of the audit events data json blob, if it exists.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -23,11 +23,13 @@ class DataAPIClient(BaseAPIClient):
             latest_first=None,
             earliest_for_each_object=None,
             user=None,
+            data_supplier_id=None,
     ):
 
         params = {
             "acknowledged": acknowledged,
             "audit-date": audit_date,
+            "data-supplier-id": data_supplier_id,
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -13,21 +13,23 @@ def init_app(self, app):
...
+            latest_first=None,
+            object_id=None,
+            object_type=None,
             page=None,
             per_page=None,
-            acknowledged=None,
-            object_type=None,
-            object_id=None,
-            latest_first=None,
-            earliest_for_each_object=None,
```
risicle: I've been told off for this before because technically changing argument order is an api-breaking change.

katstevens: Agreed - let's keep the existing order for now.

Wynndow: Okey dokey, all done

PR #156, opened 2018-09-10
Add method for transition-dos endpoint

Wynndow:
See [this closed PR ](https://github.com/alphagov/digitalmarketplace-api/pull/836)adding the endpoint to the API.

We have an new endpoint for transitioning DOS frameworks. This allows
the apiclient to call it.

### DIFF
--- a/tests/test_data_api_client.py
+++ b/tests/test_data_api_client.py
@@ -1621,6 +1621,24 @@ class TestFrameworkMethods(object):
             "updated_by": "me@my.mine"
         }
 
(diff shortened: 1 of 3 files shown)


tests/test_data_api_client.py
```diff
@@ -1621,6 +1621,24 @@ def test_update_framework(self, data_client, rmock):
             "updated_by": "me@my.mine"
         }
 
+    def test_tranision_dos_framework(self, data_client, rmock):
```
katstevens: Typo `tranision`

PR #157, opened 2018-09-11
Update setup.py requirements

lfdebrux:
When I rebuild the virtualenv for an app which has been updated to Flask 1.0.2, I get errors similar to:

```
digitalmarketplace-apiclient 19.0.0 has requirement Flask==0.10.1, but you'll have flask 1.0.2 which is incompatible.
```

This PR should fix this issue by pinning us to sensible version ranges for each dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,5 +28,5 @@ setup(
         'monotonic>=0.3',
         'requests<3,>=2.18.4',
         'six<2,>=1.11.0'
-    ] + ([] if has_enum else ['enum34==1.1.6'])
+    ] + ([] if has_enum else ['enum34<2,>=1.1.6'])
 )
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -24,9 +24,9 @@
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        'Flask==0.10.1',
```
benvand: Looks like we can just remove flask from the deps here

Nothing in this package uses it and it was added to the requirements file in 2015

benvand: Ignore, it's in tests. SMH

lfdebrux: If it's only in tests we can just move it to `requirements-dev` and leave it out of `setup.py`

benvand: 🙌

PR #159, opened 2018-10-11
BaseAPIClient: add a `timeout` constructor argument for controlling the timeout of connections

risicle:
Fell out of https://trello.com/c/aWnL8fAa/192-antivirus-create-catch-up-bulk-scan-script

Turns out we've never had timeouts set on the api client. This means requests can wait indefinitely. In the apps, it means they will continue to the point that the parent request times out the whole app request (usually it is the router that ends up doing this). In scripts it means they can hang pretty much indefinitely.

How do people feel about 45s as an initial default? I've made this a constructor argument instead of a pseudo-constant (a la `RETRIES`) because it feels we're going to want to adjust this more often.

Would people like me to add an `init_app` implementation that can pull this value from app config?

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '19.5.0'
+__version__ = '19.6.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


risicle:
Based on further discussion, we think it is probably best to use a lower (15s?) "connection" timeout but leave the "read" timeout at a high value like 45s.

PR #163, opened 2018-10-23
Adds method for accessing the endpoint

jonodrew:
This commit adds a method to the apiclient that allows a user to remove a declaration on a supplier_framework. It's implemented by calling a pre-existing method to DRY the code out as much as possible.

[Ticket](https://trello.com/c/xhq2Vy7l/211-update-apiclient-with-method-for-removing-supplierframework-declarations)

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -183,6 +183,14 @@ class DataAPIClient(BaseAPIClient):
             user=user
         )
 
+    def remove_supplier_declaration(self, supplier_id, framework_slug, user):
+        return self._post_with_updated_by(
+            "/suppliers/{}/frameworks/{}/declaration".format(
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -183,6 +183,11 @@ def set_supplier_declaration(self, supplier_id, framework_slug, declaration, use
             user=user
         )
 
+    def remove_supplier_declaration(self, supplier_id, framework_slug, user):
+        return self.set_supplier_declaration(
+            supplier_id, framework_slug, declaration={}, user=user
+        )
```
katstevens: This calls a different view - PUT instead of POST. We should be able to do something like

`return self._post_with_updated_by(<the same url>, data={}, user=user)`

PR #164, opened 2018-11-26
Support Python 3.6 only

lfdebrux:
Add python_requires specification to setup.py that explicitly declares which versions of Python this library supports/is tested on.

Part of https://trello.com/c/wWFsv1SU/230-be-clear-in-our-libraries-what-versions-of-python-we-support


lfdebrux:
> Do we care about version bumps for these?

Added

PR #168, opened 2018-12-31
Remove six library

jonodrew:
This PR is to remove uses of the `six` library. I've replaced them with their python 3 equivalents. I've also updated `setup.py` 

[Ticket](https://trello.com/c/AdmER908/149-remove-python2-deps-from-apiclient)


benvand:
Looks good but needs a version bump 👍

jonodrew:
> Looks good but needs a version bump 👍

Done, good catch thank you!

PR #178, opened 2019-01-16
Add name keyword to find_suppliers

katstevens:
Trello: https://trello.com/c/aSOduJAQ/319-admin-search-for-registered-name-as-well-as-supplier-name

Allow the apps to search the API by supplier name or registered name. See https://github.com/alphagov/digitalmarketplace-api/pull/859 for details.

Tested this locally with the Admin FE.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '19.12.0'
+__version__ = '19.13.0'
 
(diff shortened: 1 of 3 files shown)


katstevens:
I had the same problem! Naming things remains the toughest issue in software development 😹 I can't think of anything better that isn't going to be more confusing.

PR #185, opened 2019-01-21
AuditTypes: add update_selection_questions 

risicle:
https://trello.com/c/rMwWZyBv

What do people think? The kooky audit types around this area rather show their age, but this seems to fit best with them?


katstevens:
I think I'd prefer to go descriptive rather than consistent. `update_declaration_answers`?

risicle:
Um, yeah? Guess so.

PR #205, opened 2019-04-17
find_framework_suppliers default change

agalamatis:
https://trello.com/c/N6rdL6nQ/416-findframeworksuppliers-api-view-and-api-client-method-differ-on-withdeclarations-param

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Records breaking changes from major version bumps
 
+## 20.0.0
+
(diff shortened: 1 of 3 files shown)


katstevens:
So I've checked the Admin FE and our only usage there explicitly calls `with_declarations=False`, so that's fine.

In the scripts repo however, there will be some changed behaviour in a handful of the scripts - we'll now be including declaration info in the response (so it'll be much larger!). This is fine but I think it probably warrants a major bump - `20.0.0` and a line in the `CHANGLOG.md`.

PR #214, opened 2019-06-04
BaseAPIClient: make writeable attributes private

risicle:
God knows where the ticket for this is if it ever had one - it's been kicking around for a while.

This is to encourage immutabile use of api clients after construction/init_app, which should provide more threadsafety.

### DIFF
--- a/dmapiclient/base.py
+++ b/dmapiclient/base.py
@@ -49,17 +49,54 @@ def make_iter_method(method_name, *model_names):
     return iter_method
 
 
+class classproperty(object):
+    def __init__(self, getter):
+        self.getter = getter
+
+    def __get__(self, instance, owner):
+        return self.getter(owner)
+
+
 class BaseAPIClient(object):
-    RETRIES = 5
-    RETRIES_BACKOFF_FACTOR = 0.3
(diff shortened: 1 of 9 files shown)


dmapiclient/base.py
```diff
@@ -49,17 +49,58 @@ def iter_method(self, *args, **kwargs):
...
+    @property
+    def enabled(self):
+        return self._enabled
+
+    @property
+    def timeout(self):
+        return self._timeout
+
+
+    __slots__ = ("_base_url", "_auth_token", "_enabled", "_timeout",)
```
katstevens: This might be a dumb question: what is this used for?

risicle: Giving a python class `__slots__` tells python not to give it a dictionary for implementing its attributes, instead giving it a fixed selection of attributes. Basically, "don't let people assign attributes on an instance other than these listed"

katstevens: Aha, thanks! Every day is a school day.

risicle: hmm unfortunately it seems to make `mock.patch.object` very unhappy.

PR #219, opened 2019-06-19
DataAPIClient: add set_supplier_framework_allow_declaration_reuse

risicle:
https://trello.com/c/V7cUDIRW

All for the purpose of being able to _set_ this flag, yup, need a new apiclient method.


risicle:
We are in many places, just not here. Yet.

PR #220, opened 2019-07-19
Add parameter to allow finding audit events by draft service ID

lfdebrux:
Trello: https://trello.com/c/5MAl7BHa/607-auditevents-allow-draftservices-as-an-object-id

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -24,12 +24,15 @@ class DataAPIClient(BaseAPIClient):
             earliest_for_each_object=None,
             user=None,
             data_supplier_id=None,
+            *,
+            data_draft_service_id=None,
     ):
 
         params = {
             "acknowledged": acknowledged,
             "audit-date": audit_date,
             "data-supplier-id": data_supplier_id,
+            "data-draft-service-id": data_draft_service_id,
             "earliest_for_each_object": earliest_for_each_object,
             "latest_first": latest_first,
             "object-id": object_id,
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '20.2.0'
+__version__ = '20.3.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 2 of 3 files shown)


dmapiclient/data.py
```diff
@@ -24,12 +24,15 @@ def find_audit_events(
             earliest_for_each_object=None,
             user=None,
             data_supplier_id=None,
+            *,
```
katstevens: `*`? What else are we expecting?

lfdebrux: I though that `data_draft_id` should be a [keyword-only argument](https://www.python.org/dev/peps/pep-3102/)

katstevens: You could say that about all the others too (breaking changes notwithstanding) 😹 I feel like this has cropped up before but looking through old PRs I can't see it, so I might be imagining it.

katstevens:
Could we hold off on merging this until we have the new index put on the API? That way we won't (are less likely to?) use it prematurely and bring down our DB...

lfdebrux:
> Could we hold off on merging this until we have the new index put on the API? That way we won't (are less likely to?) use it prematurely and bring down our DB...

I'm happy to freeze this PR, for reference, I did test this against this against production and it did complete the request, although it took a little while.

dmapiclient/data.py
lfdebrux: Yeah, I would've put it in earlier but it would have been a breaking change.

> I feel like this has cropped up before

I've used this before here https://github.com/alphagov/digitalmarketplace-scripts/pull/388, is this the PR you meant?

katstevens: That's the one, well remembered! 

I don't think we use this elsewhere in the API client yet. Maybe we should? If so, we should be consistent with it. Perhaps a similar to our typing approach in `dmutils` - if it's a new method (or an already breaking change to an old method), start using it? Document it somewhere?

lfdebrux: I think we should use it in our api client, the reason the keyword only parameters feature was added was to be able to prevent breaking changes in apis, which is obviously useful here. I'm not sure there's any point limiting it to only breaking changes though?

PR #223, opened 2019-09-13
add more client_wait_for_response, wait_for_index options

risicle:
When attempting to implement https://trello.com/c/LABe8BKA it would be helpful to add these options to the actual api calls that the admin view in question uses in the first place.

So add `wait_for_index` option to `DataApiClient.update_service_status()`.

Also since the api's view corresponding to `update_service_status` can also _delete_ items from the index, using `SearchAPIClient.delete()`, we need to add the `client_wait_for_response` argument to _it_.

### DIFF
--- a/dmapiclient/search.py
+++ b/dmapiclient/search.py
@@ -112,11 +112,11 @@ class SearchAPIClient(BaseAPIClient):
         url = '/{}/{}/{}'.format(index_name, doc_type, object_id)
         return self._put(url, data={'document': serialized_object}, client_wait_for_response=client_wait_for_response)
 
-    def delete(self, index, service_id):
+    def delete(self, index, service_id, *, client_wait_for_response: bool = True):
         url = self._url(index, service_id)
 
         try:
-            return self._delete(url)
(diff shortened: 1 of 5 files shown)


dmapiclient/search.py
```diff
@@ -112,11 +112,11 @@ def index(
         url = '/{}/{}/{}'.format(index_name, doc_type, object_id)
         return self._put(url, data={'document': serialized_object}, client_wait_for_response=client_wait_for_response)
 
-    def delete(self, index, service_id):
+    def delete(self, index, service_id, *, client_wait_for_response: bool = True):
```
lfdebrux: Is there any reason to have this called `client_wait_for_response` rather than `wait_for_index` like `update_service_status`?

risicle: Yes, the distinction is that the argument _here_, like the one above it on `index()` is an instruction to the apiclient itself. `wait_for_index` is an instruction to the `api` to tell _it_ to use a `client_wait_for_response` argument on _its_ apiclient when _it_ is doing the daisychained call to the search api.

PR #227, opened 2020-01-15
DataAPIClient.update_service(): add page_questions argument and tests

risicle:
To aid https://trello.com/c/ShyReqk5

Analogous to same argument on `.update_brief()` method, this has been supported by the API for a long time but never exposed by the apiclient.


katstevens:
Isn't it there on `update_draft_service`?

risicle:
God _damn_ it.

PR #230, opened 2020-05-14
New method for draft services by framework endpoint

katstevens:
https://trello.com/c/kCUjDnH4/48-3-3-jenkins-job-for-publish-services

API client method (and iterable method) for the new endpoint introduced in https://github.com/alphagov/digitalmarketplace-api/pull/1030.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -560,6 +560,18 @@ class DataAPIClient(BaseAPIClient):
     find_draft_services_iter = make_iter_method('find_draft_services', 'services')
     find_draft_services_iter.__name__ = str("find_draft_services_iter")
 
+    def find_draft_services_by_framework(self, framework_slug, page=None, status=None, supplier_id=None):
+        params = {
+            'page': page,
+            'status': status,
+            'supplier_id': supplier_id
+        }
+
+        return self._get('/draft-services/framework/{}'.format(framework_slug), params=params)
+
+    find_draft_services_by_framework_iter = make_iter_method('find_draft_services_by_framework', 'services')
+    find_draft_services_by_framework_iter.__name__ = str("find_draft_services_by_framework_iter")
+
     def get_draft_service(self, draft_id):
         return self._get(
             "/draft-services/{}".format(draft_id)
--- a/tests/test_base_api_client.py
+++ b/tests/test_base_api_client.py
@@ -38,7 +38,7 @@ def _empty_context_manager():
 class TestBaseApiClient(object):
     def _from_httplib_response_mock(self, status, response_data=None):
         response_mock = mock.Mock(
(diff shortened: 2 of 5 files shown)


katstevens:
Not sure why the base client tests are suddenly failing! I'm having a look now

tests/test_base_api_client.py
```diff
@@ -38,7 +38,7 @@ def _empty_context_manager():
 class TestBaseApiClient(object):
     def _from_httplib_response_mock(self, status, response_data=None):
         response_mock = mock.Mock(
-            status=status, headers={}, spec=['get_redirect_location', 'getheader', 'read', 'reason']
+            status=status, headers={}, spec=['get_redirect_location', 'getheader', 'read', 'reason', 'drain_conn']
```
lfdebrux, line 41: Why is this needed now when it wasn't before?

dmapiclient/data.py
```diff
@@ -560,6 +560,18 @@ def find_draft_services(self, supplier_id, service_id=None, framework=None):
     find_draft_services_iter = make_iter_method('find_draft_services', 'services')
     find_draft_services_iter.__name__ = str("find_draft_services_iter")
 
+    def find_draft_services_by_framework(self, framework_slug, page=None, status=None, supplier_id=None):
+        params = {
+            'page': page,
+            'status': status,
+            'supplier_id': supplier_id
+        }
```
lfdebrux, line 568: In other endpoints we only add stuff to the dictionary if it is not None, does it matter either way?

lfdebrux:
Crazy idea, but could we make it so that we only have the one `find_draft_services(_iter)` method, and it uses a different endpoint depending on the params?

tests/test_base_api_client.py
katstevens: Good question - I spent quite a while trying to figure this out yesterday with no definitive answer. My best guess is that the pinned range for `requests` on this repo allowed an updated version that introduced this extra method (which we weren't previously stubbing out on our mock).

dmapiclient/data.py
katstevens: I did think about that at first, but for this new method I followed the style for paginated endpoints - see https://github.com/alphagov/digitalmarketplace-apiclient/blob/master/dmapiclient/data.py#L815. I don't think it makes a huge difference.

katstevens:
> Crazy idea, but could we make it so that we only have the one `find_draft_services(_iter)` method, and it uses a different endpoint depending on the params?

I think that would introduce a high risk of using the wrong one 😹

tests/test_base_api_client.py
lfdebrux: Interesting. This might be something pip-tools/Dependabot could help with.

lfdebrux: A problem for another day though.

PR #231, opened 2020-06-10
use POST email address methods

gidsg:
This updates the API client to use the methods introduced in https://github.com/alphagov/digitalmarketplace-api/pull/1037

### DIFF
--- a/tests/test_data_api_client.py
+++ b/tests/test_data_api_client.py
@@ -628,6 +628,7 @@ class TestEmailVaildForAdminMethod(object):
             status_code=200
         )
         result = data_client.email_is_valid_for_admin_user('kev@gov.uk')
+        assert rmock.last_request.json() == {"emailAddress": "kev@gov.uk"}
         assert rmock.called
         assert result is True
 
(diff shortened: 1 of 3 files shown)


lfdebrux:
### My review checklist

I've:

- [ ] looked at the code
- [ ] tested locally
- [ ] run functional tests locally

tests/test_data_api_client.py
```diff
@@ -632,8 +632,8 @@ def test_email_address_with_valid_admin_domain_is_true(self, data_client, rmock)
         assert result is True
 
     def test_email_address_with_invalid_admin_domain_is_false(self, data_client, rmock):
-        rmock.get(
-            "http://baseurl/users/valid-admin-email?email_address=kev%40not-gov.uk",
+        rmock.post(
+            "http://baseurl/users/valid-admin-email",
             json={"valid": False},
             status_code=200
         )
```
katstevens, line 639: For POST requests, we can also assert the JSON from the last call (see line 558 for an example):

```
assert rmock.last_request.json() == {
    "emailAddress": True
}
```

gidsg: great suggestion @katstevens have added these assertions in 9b3ebff.

katstevens: Great 👍 (could we add for the `test_is_email_address_with_valid_buyer_domain_(true|false)` tests as well?)

gidsg: good spot, missed those. Added those now in 55c6c37.

lfdebrux:
Functional tests passed for me.

PR #235, opened 2020-09-22
Add default `updated_by` property to BaseApiClient

lfdebrux:
This is mainly meant as a quality-of-life improvement for devs using the API client shell in digitalmarketplace-scripts; instead of having to remember to add `updater` or `user` to each method call and typing their email address every time, they can now set it once when constructing the API client object. It might also help with simplifying writing scripts.

This commit adds a new property to BaseApiClient, `_user`, which can be `None` or a string. If it is a string this will be used for `updated_by` when using a method that makes a change using the API. The caller can still use the user argument for the method call if they want to, and this will supersede the value of the property, but now they don't have to. If they haven't provided the argument in the method call or set the property, a ValueError will be raised.

This change should be backwards-compatible.

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -3,7 +3,7 @@
     "files": null,
     "lines": null
   },
-  "generated_at": "2020-08-28T14:37:56Z",
+  "generated_at": "2020-09-21T16:39:08Z",
   "plugins_used": [
     {
       "name": "AWSKeyDetector"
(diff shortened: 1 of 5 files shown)


bjgill (approved):
Nice.

I've been nursing an ambition to try and get the user's email address from their git config, and this will make that significantly easier to plumb in.

:shipit:

lfdebrux:
> Nice.
> 
> I've been nursing an ambition to try and get the user's email address from their git config, and this will make that significantly easier to plumb in.
> 
> :shipit:

Ah, maybe it was you I stole the idea from... https://github.com/alphagov/digitalmarketplace-scripts/compare/ldeb-add-updated-by-helpers Feel free to finish it!

PR #240, opened 2020-12-18
Document responses that are sometimes paginated

bjgill:
Some API endpoints are paginated and some are not. But there also exist three that may be paginated depending on the request parameters. This last category is rather confusing, so document them.


alex9smith (approved):
Thanks for digging into this!

PR #242, opened 2021-01-08
Deprecate paginated methods with an alternative

bjgill:
The paginated methods are dangerous. It's not obvious without reading the source that they are paginated, so it's very easy to write scripts that will fail when the number of API object it's operating on becomes large. In scripts, you almost never want pagination.

Most paginated methods also have a non-paginated '_iter' alternative that is much safer. Deprecate the paginated versions. Pycharm will now warn about uses of the deprecated methods. You can still use the methods - you'll just get a warning.

We've been bitten a few times recently by accidentally using paginated API methods inappropriately. This should hopefully make that sort of mistake less common in future.

### DIFF
--- a/dmapiclient/__init__.py
+++ b/dmapiclient/__init__.py
@@ -1,4 +1,4 @@
-__version__ = '21.10.1'
+__version__ = '21.11.0'
 
 from .errors import APIError, HTTPError, InvalidResponse  # noqa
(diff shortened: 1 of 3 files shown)


lfdebrux (approved):
Bold, I like this!

bjgill:
Thanks. I had a look to see if I could do it more elegantly with decorators or metaprogramming. Unfortunately, if I do anything more clever, PyCharm stops recognising it. So simple and slightly boilerplate-y it is.

PR #245, opened 2021-02-23
Add delete_buyer_email_domain audit event type

bjgill:
https://govuk.zendesk.com/agent/tickets/4475776

We're going to add a new API endpoint to allow us to delete domains from the buyer allow list. So we need a new audit event type.


gidsg (approved):
🔍 👍

PR #246, opened 2021-02-23
Support new API endpoints

bjgill:
As added by https://github.com/alphagov/digitalmarketplace-api/pull/1089 and https://github.com/alphagov/digitalmarketplace-api/pull/1094


bjgill:
I'm not at all convinced that the tests are adding much, but I'm being consistent.

alex9smith (approved):
👍

PR #249, opened 2021-04-09
Add type hint checking

bjgill:
Use mypy to check the types in this library. Then mark the library as having type hints so consumers can use them.

Note that the type hinting is thus far rather sparse, so this will not yet be very useful to consumers.

Copied very heavily from https://github.com/alphagov/digitalmarketplace-utils/pull/606

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -9,3 +9,4 @@ def test_mypy(c):
 
 
 ns.add_task(test_mypy)
+ns["test"].pre.insert(-1, test_mypy)
--- a/tasks.py
+++ b/tasks.py
@@ -5,7 +5,7 @@ from dmdevtools.invoke_tasks import library_tasks as ns
 
 @task(ns["virtualenv"], ns["requirements_dev"])
 def test_mypy(c):
-    c.run("mypy dmutils/")
+    c.run("mypy")
 
 
 ns.add_task(test_mypy)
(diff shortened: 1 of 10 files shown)


tasks.py
```diff
@@ -1 +1,12 @@
+from invoke import task
+
 from dmdevtools.invoke_tasks import library_tasks as ns
+
+
+@task(ns["virtualenv"], ns["requirements_dev"])
+def test_mypy(c):
+    c.run("mypy dmutils/")
```
alex9smith, line 8:
```suggestion
    c.run("mypy dmapiclient/")
```
I _think_ this is the only reason why the checks are failing - mypy seems happy on my machine

lfdebrux: Might be worth instead adding a `mypy.ini` file

```
[mypy]
files = dmapiclient
```

Then running `mypy` should just do the right thing.
```

bjgill: Huh. I didn't know about that. We already have a `mypy.ini`, so that sounds most useful

bjgill: Fixed and rebased against main

bjgill:
Oh dear - tests broken for an unrelated reason that's going to hit an number of other repos. I've created a ticket: https://trello.com/c/vRE4IeaF/2206-fix-github-actions-ci-for-repos-that-dont-have-pip-tools

lfdebrux (approved):

PR #251, opened 2021-04-21
Update requirements files to be managed with pip-tools

lfdebrux:
Ticket: https://trello.com/c/vRE4IeaF/2206-fix-github-actions-ci-for-repos-that-dont-have-pip-tools

digitalmarketplace-developer-tools>=1.0.0 requires the requirements files to be managed with pip-compile. This commit removes the requirements.txt file (which did nothing basically) and adds a requirements-dev.in file so that requirements-dev.txt can include the full set of requirements for working on this library.


bjgill (approved):
Ah. So I'd misunderstood the problem. It wasn't that we needed to add pip-tools as a dev dependency. It's that we needed to add digitalmarketplace-developer-tools as a dev dependency. Makes sense.

:tada:

bjgill:
And deleting the almost-empty requirements.txt file, which seems like the main problem for pip-sync

lfdebrux:
> Ah. So I'd misunderstood the problem. It wasn't that we needed to add pip-tools as a dev dependency. It's that we needed to add digitalmarketplace-developer-tools as a dev dependency. Makes sense.

Actually it's neither... we could remove digitalmarketplace-developer-tools as a dev dependency if we wanted too.

lfdebrux:
I think it's just that pip-sync doesn't like requirements files it didn't generate itself; the error message is mightily unhelpful though :/

lfdebrux:
> Actually it's neither... we could remove digitalmarketplace-developer-tools as a dev dependency if we wanted too.

Thinking about it, I think it makes more sense not to have dmdevtools in the venv (it's not actually needed there currently), so I might remove dmdevtools from the requirements file, if you're happy with that approach.

bjgill:
> > Actually it's neither... we could remove digitalmarketplace-developer-tools as a dev dependency if we wanted too.
> 
> Thinking about it, I think it makes more sense not to have dmdevtools in the venv (it's not actually needed there currently), so I might remove dmdevtools from the requirements file, if you're happy with that approach.

Makes sense. It took me a few tries to understand what was going on here, and I agree we don't need it in the venv

bjgill (approved):
:shipit:

PR #253, opened 2021-06-15
Add support for status_date_filters to find_briefs

bjgill:
https://trello.com/c/FGm8bi13/3-2-suppliers-may-be-missing-out-on-opportunities-due-to-misleading-banner

There are 18 possible status_date_filters, so we don't want to add them all individually. Instead allow users to provide a dictionary and let the API work out whether it's correct or not.

This should make a spot of analysis I'm trying to do a bit easier.

### DIFF
--- a/dmapiclient/data.py
+++ b/dmapiclient/data.py
@@ -1,6 +1,7 @@
 from __future__ import unicode_literals
 
 import warnings
+from typing import Dict
 
 from .audit import AuditTypes
 from .base import BaseAPIClient, logger, make_iter_method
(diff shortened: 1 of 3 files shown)


dmapiclient/data.py
```diff
@@ -907,31 +908,40 @@ def get_brief(self, brief_id):
...
     def find_briefs(
         self, user_id=None, status=None, framework=None, lot=None, page=None, human=None, with_users=None,
-        with_clarification_questions=None, closed_on=None, withdrawn_on=None, cancelled_on=None, unsuccessful_on=None
+        with_clarification_questions=None, closed_on=None, withdrawn_on=None, cancelled_on=None, unsuccessful_on=None,
+        status_date_filters: Dict[str, str] = None,
     ):
         """
         The response will be paginated unless you provide user_id.
+
+        :param status_date_filters: contains additional status date filters. See the API for permitted keys.
```
gidsg, line 917: Can we be any more specific about where these are listed/defined?

PR #254, opened 2021-06-17
Add support for sort_by to find_audit_events

bjgill:
This was added to the API recently: https://github.com/alphagov/digitalmarketplace-api/pull/1147. Add support to the API client.

Should make the service edit report easier.


alex9smith (approved):
👍

